PR #7, opened 2015-03-31
Delete setup.cfg

The-Compiler:
Just noticed this has the same issue as in https://github.com/nicoddemus/pytest-faulthandler/pull/2

```
$ python --version       
Python 3.4.3
$ pip install pytest-mock
Collecting pytest-mock
  Downloading pytest_mock-0.4.1-py2.py3-none-any.whl
Collecting mock (from pytest-mock)
  Downloading mock-1.0.1.tar.gz (818kB)
...
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,0 @@
-[bdist_wheel]
-universal = 1
-


nicoddemus:
Hi @The-Compiler,

Yep, I noticed that as well before coming to work this morning, but I wasn't too worried because at least it doesn't give an error like pytest-faulthandler did.

Thanks again for the interest and submitting PRs for those issues. I will make a new release later today without a wheels package. 

Cheers!

PR #11, opened 2015-05-03
Add Mock/MagicMock to MockFixture class for instantiation

marcwebbie:
Instantiate Mocks and MagicMocks using mocker.

Example use case:

```python
def test_my_feature_with_pytest_mock(mocker):
    ret = [mocker.Mock(return_value=True), mocker.Mock(return_value=True)]
    mocker.patch('mylib.func', side_effect=ret)
    ...
```


nicoddemus:
Hey @marcwebbie, many thanks for this! :smile:

PR #12, opened 2015-06-01
Added methods spy and stub to mocker fixture

fogo:


test_pytest_mock.py
```diff
@@ -136,3 +136,26 @@ def test_mocker_has_mock_class_as_attribute_for_instantiation():
...
+
+
+def test_mocker_spy(mocker):
+    class Foo(object):
+
+        def bar(self, arg):
+            return arg * 2
+
+    foo = Foo()
+    spy = mocker.spy(foo, 'Bar')
```
nicoddemus: You mean `'bar'` here, right? :)

nicoddemus:
Could you also please announce on the README about the new `spy` and `stub` methods? Include a simple usage of each one as well, and add a note below each one like *New in version 0.6*.

test_pytest_mock.py
fogo: Ops, that's what happens when you push in a hurry because of a meeting.

nicoddemus:
Thanks! :smile:

PR #13, opened 2015-07-06
Now possible to spy methods in class, static and class methods

fogo:
Also improved docs, made more compliant to reStructuredText expectations.


nicoddemus:
Thanks! I will make a release later today! :smile:

PR #14, opened 2015-07-06
More spy

nicoddemus:


nicoddemus:
@fogo, couldn't make this work with the case of patching `Class.method` and calling instances... let's discuss this in more details when we can.

nicoddemus:
Figured out! :sweat_smile:

PR #16, opened 2015-07-15
Update README.rst with contextmanager warning

tigarmo:
It's not supported because it's not needed, but the docs can at least make the disclaimer!


nicoddemus:
Thanks! :smile:

PR #17, opened 2015-07-29
Expose mock.ANY in mocker fixture

tigarmo:
It's handy for when you want to check _some_ parameters in a call list but not necessarily all of them.

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -18,6 +18,7 @@ class MockFixture(object):
 
     Mock = mock_module.Mock
     MagicMock = mock_module.MagicMock
+    ANY = mock_module.ANY
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks! :smile:

The-Compiler:
Commenting here for the lack of a better place - the release notes on GitHub say:

> mock.ANY is also accessible from the mocker fixture to make (#17)

It seems something is missing there :wink:

nicoddemus:
You mean the message is incomplete?

The-Compiler:
Yes, or at least I can't figure out what it means.

nicoddemus:
Oops! :sweat_smile: 

Fixed:

> mock.ANY is also accessible from the mocker fixture (#17)

Thanks for the heads up! :smile:

PR #18, opened 2015-09-29
Use PEP-0426 environment markers to build universal wheels

rouge8:
PEP-0426 defines [environment markers][] to allow wheels to conditionally install dependencies. This won't work with really old versions of setuptools/pip, but it works with pip 1.5.6 and newer.

[environment markers]: https://www.python.org/dev/peps/pep-0426/#environment-markers

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal=1
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,5 @@
-import sys
-
 from setuptools import setup
 
-
-dependencies = ['pytest>=2.4']
-if sys.version_info < (3, 3):
-    dependencies.append('mock')
 
 setup(
     name='pytest-mock',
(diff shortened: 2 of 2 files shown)


nicoddemus:
Thanks @rouge8, I did not know about environment markers! :smile:

nicoddemus:
Hmm the CI broke and I couldn't get it to work locally too... do you know what's the problem?

nicoddemus:
Hmm it seems that the statement doesn't support the `<` operator... I could get this to work:
```
extras_require={
        ':python_version=="3.2" or python_version=="2.6" or '
        'python_version=="2.7"': ['mock'],
    },
```

rouge8:
Hmm, succeeds for me... I wonder what version of pip/setuptools introduced support for comparisons...

rouge8:
Aha, setuptools 17.1: https://bitbucket.org/pypa/setuptools/issues/380/environment-markers-dont-support-etc That's pretty recent, so #19 is probably better.

nicoddemus:
That explains it! :smile:

Thanks! I will take the opportunity to add `py35` and AppVeyor.

Cheers,
:beers:

PR #24, opened 2015-11-11
Stab at implementing resetall(). See #23

### LINKED ISSUE #23: Add mocker.resetall()
Having a `mocker.resetall()` method, similar to `mocker.stopall()`, would be very handy. Most of the time if I mock things, it's because I'm testing a side-effecting function which cannot be blackboxed effectively, and will want to walk through multiple code paths while checking that various callees were/were not called as appropriate. This commonly involves resetting the mocks, so not having to reset them individually would greatly increase the usefulness of `mocker`.

mathrick:
I poked around and figured out some kind of approach that works to fix #23 in my limited testing. It might very well be completely wrong, I don't understand the internals well enough to know what I'm doing.

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -24,6 +24,13 @@ class MockFixture(object):
         self._patches = []  # list of mock._patch objects
         self.patch = self._Patcher(self._patches)
 
+    def resetall(self):
+        """
+        Call reset_mock() on all patchers started by this fixture.
+        """
+        for p in self._patches:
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks @mathrick, hopefully I will have time to review and merge or at least use this as a starting point. :smile:

nicoddemus:
`0.9.0` is now out with this.

Thanks again for the PR and sorry for taking so long. :smile:

PR #25, opened 2015-11-18
Resetall

nicoddemus:
Supersedes #24

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+0.9.0
+-----
+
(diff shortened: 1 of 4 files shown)


nicoddemus:
@mathrick I opted to keep a list of `MagicMock` objects explicitly instead, to use the public API as `get_original` is not part of the public API it seems.

mathrick:
Yeah, I thought about doing that after I created the PR, and it's a cleaner and stabler approach. I just got caught up in trying to find a "built-in" way and it didn't occur to me back then.

nicoddemus:
No problem, thanks again! :smile:

PR #27, opened 2016-01-27
Wrap assert methods of mock module to clean up traceback

### LINKED ISSUE #26: stub.assert_called_once_with has distracting reporting
`assert_called_once_with` and the other `assert_` functions show in the traceback, which is very unhelpful. `__tracebackhide__` would help, but this is is a system lib :).

## Test
```
def test_stub(mocker):
    stub = mocker.stub()
    stub.assert_called_once_with('foo', 'baz')
```
## Output:
```
================================================================================== FAILURES ==================================================================================
_________________________________________________________________________________ test_stub __________________________________________________________________________________

mocker = <pytest_mock.MockFixture object at 0x7f52cb320cc0>

    def test_stub(mocker):
        stub = mocker.stub()
>       stub.assert_called_once_with('foo', 'baz')

mocker     = <pytest_mock.MockFixture object at 0x7f52cb320cc0>
stub       = <MagicMock spec='function' id='139993573101520'>

tests/ston/test_utils.py:21:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

_mock_self = <MagicMock spec='function' id='139993573101520'>, args = ('foo', 'baz'), kwargs = {}, self = <MagicMock spec='function' id='139993573101520'>
msg = "Expected 'mock' to be called once. Called 0 times."

    def assert_called_once_with(_mock_self, *args, **kwargs):
        """assert that the mock was called exactly once and with the specified
            arguments."""
        self = _mock_self
        if not self.call_count == 1:
            msg = ("Expected '%s' to be called once. Called %s times." %
                   (self._mock_name or 'mock', self.call_count))
>           raise AssertionError(msg)
E           AssertionError: Expected 'mock' to be called once. Called 0 times.

_mock_self = <MagicMock spec='function' id='139993573101520'>
args       = ('foo', 'baz')
kwargs     = {}
msg        = "Expected 'mock' to be called once. Called 0 times."
self       = <MagicMock spec='function' id='139993573101520'>

/usr/lib/python3.4/unittest/mock.py:781: AssertionError
```
## Expected Output:
```
================================================================================== FAILURES ==================================================================================
_________________________________________________________________________________ test_stub __________________________________________________________________________________

mocker = <pytest_mock.MockFixture object at 0x7f07b55ed128>

    def test_stub(mocker):
        stub = mocker.stub()
>       stub.assert_called_once_with('foo', 'baz')
E       AssertionError: Expected 'mock' to be called once. Called 0 times.

mocker     = <pytest_mock.MockFixture object at 0x7f07b55ed128>
stub       = <MagicMock spec='function' id='139671084389320'>

tests/ston/test_utils.py:21: AssertionError
```

Chronial:
Fixes #26

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -2,6 +2,7 @@ import inspect
 import sys
 
 import pytest
+from distutils.util import strtobool
 
 if sys.version_info >= (3, 3): # pragma: no cover
     import unittest.mock as mock_module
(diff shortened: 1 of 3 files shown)


Chronial:
This is proving to be more difficult than expected ^^.

Chronial:
There we go, working now.

pytest_mock.py
```diff
@@ -137,3 +137,84 @@ def mock(mocker):
...
+
+
+def unwrap_assert_methods():
+    for patcher in _mock_module_patches:
+        patcher.stop()
+    _mock_module_patches[:] = []
+    _mock_module_originals.clear()
+
+
+def pytest_configure(config):
```
nicoddemus: I think for completeness we should also implement `pytest_unconfigure` to undo all the patches.

Chronial: Yes, I had that – but then the tests break, because the tests actually run a second pytest in the same process. That pytest will also call unconfigure, and the patching will be gone for all following tests.

But see line 209 – that does exactly what you're asking for.

nicoddemus: I'm a little uneasy of doing this wrapping automatically for everyone which uses the plugin... there are a tons of test suites, some of them doing already some deep dark magic on their on, who knows what this might break? I agree with you it is unlikely, but I think we should add an option to disable this. How about an ini option:

```ini
[pytest]
mock_traceback_monkeypatch = false
```

The default would be `true`, but users can disable it at will at least. What do you think?

test_pytest_mock.py
```diff
@@ -246,3 +249,63 @@ def bar(arg):
...
+        yield
+    except AssertionError:
+        traceback = py.code.ExceptionInfo().traceback
+        crashentry = traceback.getcrashentry()
+        assert crashentry.path == __file__
+    else:
+        raise AssertionError("DID NOT RAISE")
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 4) and sys.version_info < (3, 5),
```
nicoddemus: The logic can be just: `sys.version_info[:2] == (3, 4)` :smile:

pytest_mock.py
```diff
@@ -137,3 +137,84 @@ def mock(mocker):
...
+    __tracebackhide__ = True
+    try:
+        method(*args, **kwargs)
+    except AssertionError as e:
+        raise AssertionError(*e.args)
+
+
+def wrap_assert_not_called(*args, **kwargs):
+    __tracebackhide__ = True
+    assert_wrapper(_mock_module_originals["assert_not_called"],
```
nicoddemus: Couldn't you just call the wrapped method directly, and let the `AssertionError` propagate? Like this:

```python
def wrap_assert_not_called(*args, **kwargs):
    __tracebackhide__ = True
    _mock_module_originals["assert_not_called"], *args, **kwargs)
```

pytest_mock.py
Chronial: Yes, I agree – it's quite agressive and I felt the pain of this while writing it. But as I can't see any other way to make sure that the patching is there when somebody uses mock in a test, I think a config option would probably be a good way to handle problems. I'll add one.

nicoddemus: Thanks! :smile:

nicoddemus: > But see line 209 – that does exactly what you're asking for.

Oops, missed that, thanks!

nicoddemus:
Hey thanks a lot for the PR! The code looks great!

Other than my other comments, the only thing left to do is mention this new feature in the `README`.

pytest_mock.py
Chronial: That would defeat the whole purpose :). Then the traceback would look exactly like it did before – the wrapper is hidden, but the `assert_not_called` frame is displayed.

test_pytest_mock.py
Chronial: Well yes, that sounds simpler :smile:

pytest_mock.py
nicoddemus: Oh you are right, I was under the impression that having a local variable named `__tracebackhide__` would strip all frames below and including that frame, but it only hides the frame containing the variable so you have to raise a new exception at that point. Got it, thanks! :smile:

Chronial:
done :smile:

nicoddemus:
Thanks a lot! I will merge this tonight and make a new release.  :smile:

The-Compiler:
:+1:

nicoddemus:
Released 0.10.0 :smile:

Thanks again!

PR #32, opened 2016-02-20
Add PropertyMock to mocker

satyrius:
I missed PropertyMock in mocker https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock


nicoddemus:
Thanks for the contribution! :smile:

It will be out in `0.11` in a few days.

satyrius:
Nice, thanks a lot!

nicoddemus:
Released `0.11.0`, thanks!

PR #36, opened 2016-03-08
Show detailed py.test assert introspect for call arguments

asfaltboy:
## Why do this

I really enjoy pytest's assertion introspection support, and find it invaluable when it comes to comparing large pieces of data. However, when asserting mock calls with large arguments, I found myself copy-pasting objects into [`deepdiff`][1] just to more easily see the difference.

## Implementation

This may not be a complete solution, but it's the simplest I could think of to start discussing the point.

Among other improvements, we could also check `--assert` mode option, use [a custom assertion comparison][2] and [`raise_from`][3] original mock assertion.

Please let me know what you think about this feature?

[1]: https://pypi.python.org/pypi/deepdiff
[2]: https://pytest.org/latest/assert.html#defining-your-own-assertion-comparison
[3]: https://docs.python.org/3/library/exceptions.html#built-in-exceptions

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,6 +121,16 @@ your ``pytest.ini`` file:
     [pytest]
     mock_traceback_monkeypatch = false
 
+The plugin also adds introspection information on differing call arguments when
+calling the helper methods. This features catches `AssertionError` raised in
+the method, and uses py.test's own `advanced assertions`_ to return a better
+diff.
+
+This is useful when asserting mock calls with many/nested arguments and trying
+to quickly see the difference.
+
+.. _advanced assertions: https://pytest.org/latest/assert.html
+
 
 Requirements
 ============
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -154,6 +154,11 @@ def assert_wrapper(__wrapped_mock_method__, *args, **kwargs):
     try:
         __wrapped_mock_method__(*args, **kwargs)
     except AssertionError as e:
+        __mock_self = args[0]
+        if __mock_self.call_args is not None:
+            actual_args, actual_kwargs = __mock_self.call_args
+            assert actual_args == args[1:]
+            assert actual_kwargs == kwargs
         raise AssertionError(*e.args)
 
 
@@ -189,7 +194,8 @@ def wrap_assert_any_call(*args, **kwargs):
 
 def wrap_assert_methods(config):
     """
-    Wrap assert methods of mock module so we can hide their traceback
+    Wrap assert methods of mock module so we can hide their traceback and
+    add introspection information to specified argument asserts.
     """
     # Make sure we only do this once
     if _mock_module_originals:
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -330,6 +330,25 @@ def assert_traceback():
         raise AssertionError("DID NOT RAISE")
 
 
+@contextmanager
+def assert_argument_introspection(left, right):
+    """
+    Assert detailed argument introspection is used
+    """
+    try:
+        yield
+    except AssertionError as e:
+        # this may be a bit too assuming, but seems nicer then hard-coding
+        import _pytest.assertion.util as util
+        # NOTE: we assert with either verbose or not, depending on how our own
+        #       test was run by examining sys.argv
+        verbose = any(a.startswith('-v') for a in sys.argv)
+        expected = '\n  '.join(util._compare_eq_iterable(left, right, verbose))
+        assert expected in e.msg
+    else:
+        raise AssertionError("DID NOT RAISE")
+
+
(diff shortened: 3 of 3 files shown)


nicoddemus:
Hey @asfaltboy thanks for the PR! I didn't have time yet to try it out in detail, I will do it until the end of the week

nicoddemus:
Hey @asfaltboy sorry for not replying earlier.

I like the idea in general, although I'm not sure I like the proposed UI. Perhaps we could just provide similar functions in the `mocker` fixture instead of a context manager?

```python
def test_assert_called_args_with_introspection(mocker):
    stub = mocker.stub()
    mocker.assert_called_with(stub, *args, **kwargs)
    # or perhaps
    assert mocker.called_with(stub, *args, **kwargs)
```

What do you think?

asfaltboy:
The usage UI is basically unchanged; which is the whole idea of this patch. Instead of manually dissecting arguments and calling, for example, `assert X == Y` the users can now use mock's methods (namely `assert_called_with` and `assert_called_once_with`)  and have the same nice diff as they expect when using `py.test` normally.

I'm sorry if my code is unclear, the context manager *is used ONLY in own tests* to test that the raised `AssertionError` includes a detailed diff (i.e introspection of objects). Honestly, I only used it because I saw the [`assert_traceback` function][1], and wanted to conform to project style. The actual entry point of the feature is in the `try/except` clause that already happens in [`assert_wrapper`][2].

If you prefer, we can probably as easily move the `try/except` clause around the [yield in `mocker` fixture][3]. Though, I rather like the idea of *plugging in* `pytest-mock` into an existing project (which already uses `mock`) and getting the same nice introspection with minimal work.

P.S: I was about to write this as a plugin for myself, but then I found your project and thought it fitting very much with the purpose of `pytest-mock`.

[1]: https://github.com/asfaltboy/pytest-mock/blob/feature/introspect-called-with/test_pytest_mock.py#L262
[2]: https://github.com/asfaltboy/pytest-mock/blob/feature/introspect-called-with/pytest_mock.py#L148
[3]: https://github.com/asfaltboy/pytest-mock/blob/feature/introspect-called-with/pytest_mock.py#L129

nicoddemus:
> The actual entry point of the feature is in the try/except clause that already happens in assert_wrapper.

Oh of course, sorry about that... you are right, I misunderstood the patch.

I agree that it would be nice to have this in `pytest-mock`. Are you willing to continue working on the PR? From my point of view what's missing is docs and tests which cover all functions we are currently wrapping to make sure we are not breaking any of them by accident.

asfaltboy:
Sure, I can probably do some work during this weekend.

Current tests already call all of the `mock` helper functions (as part of the `assert_traceback` tests).
I can add the other calls in separate tests, but the feature doesn't add anything we can assert in these cases, so it feels a bit repetitive.

With regards to docs, I'm thinking to add a paragraph in (or under) the ["improved reporting"][1] section.

By the way, do you feel we should add an option to toggle off "advanced introspection"? Something like:

```ini
[pytest]
mock_argument_introspection_monkeypatch = false
```

[1]: https://github.com/pytest-dev/pytest-mock#improved-reporting-of-mock-call-assertion-errors

nicoddemus:
> Current tests already call all of the mock helper functions (as part of the assert_traceback tests).

Oh you are right.

> With regards to docs, I'm thinking to add a paragraph in (or under) the "improved reporting" section.

Sounds good.

> By the way, do you feel we should add an option to toggle off "advanced introspection"?

Sounds good too. :grin: 

Thanks a lot for taking the time to work on this, I really appreciate it.

asfaltboy:
Sorry for the delay @nicoddemus. I've rebased over the current master and added docs. Unfortunately adding a toggle proved to be difficult to do cleanly and I keep running out of free time.
Please let me know if you want me to add any other changes.

blueyed:
I've just tried this, but am also seeing https://github.com/pytest-dev/pytest-mock/issues/44 with this (two tracebacks).
The diff with this is enhanced as expected though (with the second traceback).

However, the behavior changes: while you see the whole lists by default, you need to use `-vv` now to see the whole diff, which is still different from seeing the whole lists (which are easier to copy).
Given that it's actually helpful to see the original AssertionError with that information, but if I understand it correctly it should not be visible?!

asfaltboy:
@blueyed you are right of course, the behavior is different from original `mock` output. Personally, I find the enhanced output of `py.test` to better suit my needs, adding `-v` where I want to have a deeper look is easy. However, it would probably be best to make the feature toggleable for existing users, if only I had more time.

About #44, I'll leave a comment there.

blueyed:
> adding -v where I want to have a deeper look is easy

Maybe `-vv` should also include the original exception's message?

nicoddemus:
Just merged this, will deal with #44 in a separate PR. :grin: 

Thanks again @asfaltboy! :smile:

PR #41, opened 2016-05-17
Add Python 3.2 and 3.3 to test matrix

jayvdb:


nicoddemus:
Hi @jayvdb thanks for taking the time to work on this!

TBH I don't intend to support 3.2 and 3.3, as I figured 3.4+ was good enough for most people (You can see I explicitly dropped support for them in 9da34539). Do you need 3.2 and/or 3.3?

jayvdb:
I need py33 for pywikibot (currently investigating replacing nose + plugins), and I want py32 so I can use it for testing some build tools which need to work on all versions, but I could live without py32 and am getting used to maintaining my own forks of tools for working with py32.

The-Compiler:
FWIW I'd say let's add 3.3 support back as it shouldn't be too painful.

As for 3.2, a lot of other projects dropped it already (notably coverage.py, pip, virtualenv), so still trying to support that sounds like too much pain.

Then again it's @nicoddemus' call, I'm more an user of than a contributor to this plugin.

nicoddemus:
@jayvdb I see, thanks for the answer.

I agree with @The-Compiler though (in fact I was writing exactly what he wrote but he beat me to it), keeping support for 3.2 seems like a pain, even AppVeyor doesn't have that installed by default anymore.

What do you think @jayvdb? If you agree, would you please adjust your PR accordingly and add 3.3 to `setup.py`'s classifiers?

jayvdb:
Happy to amend it for py33 only. Will be an hour or before I get back to a workstation.

nicoddemus:
OK, thanks! :smile:

jayvdb:
py33 only submitted as #43 (as this PR has a branch name 'py32')

PR #42, opened 2016-05-17
spy: handle AttributeError with __getattribute__

blueyed:
I have noticed that `spy` did not work on functions that are not defined
in the class itself, but a parent class.  This patch fixes this.

### DIFF
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -212,12 +212,51 @@ def test_instance_method_by_class_spy(mocker):
 
 
 @skip_pypy
+def test_instance_method_by_subclass_spy(mocker):
+    from pytest_mock import mock_module
+
+    class Base(object):
+
+        def bar(self, arg):
+            return arg * 2
+
+    class Foo(Base):
+        pass
+
+    spy = mocker.spy(Foo, 'bar')
+    foo = Foo()
+    other = Foo()
+    assert foo.bar(arg=10) == 20
+    assert other.bar(arg=10) == 20
+    calls = [mock_module.call(foo, arg=10), mock_module.call(other, arg=10)]
+    assert spy.call_args_list == calls
+
+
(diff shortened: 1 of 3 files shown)


test_pytest_mock.py
```diff
@@ -226,6 +247,24 @@ def bar(cls, arg):
 
 
 @skip_pypy
+def test_class_method_subclass_spy(mocker):
+    class Base(object):
+
+        @classmethod
```
nicoddemus: Is this supposed to be a `classmethod`?

nicoddemus:
Thanks a lot @blueyed!

Could you please also write a new CHANGELOG entry?

The timing is great, I was planning on publishing a new release today! :smile:

test_pytest_mock.py
blueyed: Yep, it is now - I was pushing a fixup/amended commit.

blueyed:
It does not work with Python 2 apparently:

```
    @skip_pypy
    def test_static_method_subclass_spy(mocker):
        class Base(object):
    
            @staticmethod
            def bar(arg):
                return arg * 2
    
        class Foo(Base):
            pass
    
        spy = mocker.spy(Foo, 'bar')
>       assert Foo.bar(arg=10) == 20
E       TypeError: unbound method bar() must be called with Foo instance as first argument (got nothing instead)
```

(same with classmethod).

Should the tests be skipped there?

nicoddemus:
Hmmm that's a shame.

Please `xfail` them in Python 2 instead (`@pytest.mark.xfail(sys.python_version[0] == 2, reason='does not work on Python 2')` should do the trick. We can try to make it work later, but the PR as it stands is already an improvement.

blueyed:
Done.

nicoddemus:
Thanks again! :smile:

PR #46, opened 2016-05-25
Disable monkey patching "mock" when --tb is set to native

### LINKED ISSUE #44: Two tracebacks with assert_wrapper / wrap_assert_has_calls
When using `assert_has_calls` I am seeing two tracebacks, but the first one should not be there?!

```
  File "…/pytest-mock/pytest_mock.py", line 155, in assert_wrapper
  File "/usr/lib64/python3.5/unittest/mock.py", line 824, in assert_has_calls
    ) from cause
AssertionError: Calls not found.
Expected: […]
Actual: […]

During handling of the above exception, another exception occurred:

  File "…/app/tests/test_metrics.py", line 495, in test_handle_missing_values
    expected_call(end, 2, None),
  File "…/pytest-mock/pytest_mock.py", line 181, in wrap_assert_has_calls
    *args, **kwargs)
  File "…/pytest-mock/pytest_mock.py", line 157, in assert_wrapper
    raise AssertionError(*e.args)
AssertionError: Calls not found.
Expected: […]
Actual: […]
```

This is with pytest-2.9.2.dev1 and Python 3.5.1.

It looks like the `__tracebackhide__ = True` trick is not working?!

With `mock_traceback_monkeypatch=false` I only get the original traceback as expected.

nicoddemus:
The mechanism we use to hide traceback entries from the `mock` module does not work on native tracebacks anyway, so it makes sense to automatically disable it in that situation.

This also takes care of the duplicated tracebacks on Python 3.5. @blueyed could you try this out and tell what do you think?

Fix #44

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,10 +14,16 @@
   ``repr`` string as well as related assertion failure messages.
   Thanks `@jurko-gospodnetic`_ for the PR (`#40`_).
 
+* Monkey patching is automatically disabled with the ``--tb=native`` option. The underlying
+  mechanism used to suppress traceback entries from ``mock`` module does not work with that option
+  anyway plus it generates confusing messages on Python 3.5 due to exception chaining (`#44`_).
+  Thanks `@blueyed`_ for the report.
+
 .. _@jurko-gospodnetic: https://github.com/jurko-gospodnetic
 .. _@asfaltboy: https://github.com/asfaltboy
-.. _#40: https://github.com/pytest-dev/pytest-mock/issues/40
 .. _#36: https://github.com/pytest-dev/pytest-mock/issues/36
+.. _#40: https://github.com/pytest-dev/pytest-mock/issues/40
+.. _#44: https://github.com/pytest-dev/pytest-mock/issues/44
 
 1.0
 ---
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -370,7 +370,7 @@ def assert_argument_introspection(left, right):
         raise AssertionError("DID NOT RAISE")
 
 
-@pytest.mark.skipif(sys.version_info[0] == 3 and sys.version_info[1] in (3, 4),
+@pytest.mark.skipif(sys.version_info[:2] in [(3, 3), (3, 4)],
                     reason="assert_not_called not available in python 3.3 and 3.4")
 def test_assert_not_called_wrapper(mocker):
     stub = mocker.stub()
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -14,10 +14,16 @@
   ``repr`` string as well as related assertion failure messages.
   Thanks `@jurko-gospodnetic`_ for the PR (`#40`_).
 
+* Monkey patching is automatically disabled with the ``--tb=native`` option. The underlying
```
nicoddemus: Hmm I should change this to:

```
Monkey patching ``mock`` module for friendlier tracebacks is automatically disabled 
with the ``--tb=native`` option.
```

nicoddemus:
@blueyed did you get a chance of trying this out? :grin:

blueyed:
Thanks, I have not tried it yet - sorry.

Would it make sense to display a warning/info message in case it gets automatically disabled?
That would be useful especially when you try to explicitly enable / use it (and not if it's just the default).

nicoddemus:
> Would it make sense to display a warning/info message in case it gets automatically disabled?

Hmmm it does make sense, not sure how to display this warning though. How about:

```
$ py.test --tb=native
============================= test session starts =============================
platform win32 -- Python 2.7.11, pytest-2.9.0, py-1.4.31, pluggy-0.3.1
qt-api: pyqt5
mock: monkeypatching disabled because of --tb=native
rootdir: x:/projects, inifile: pytest.ini
plugins: mock-0.11.0, qt-1.3.0,
collecting...
```

blueyed:
> mock: monkeypatching disabled because of --tb=native

:+1: 

Also the PR works in general as advertised.

nicoddemus:
@The-Compiler what do you think? :grin:

CHANGELOG.rst
The-Compiler: You didn't yet :wink:

nicoddemus: Yep, thanks for the reminder. :grin:

test_pytest_mock.py
```diff
@@ -489,3 +489,21 @@ def request(cls, method, args):
...
+    """)
+    if hasattr(testdir, 'runpytest_subprocess'):
+        result = testdir.runpytest_subprocess('--tb=native')
+    else:
+        # pytest 2.7.X
+        result = testdir.runpytest('--tb=native')
+    assert result.ret == 1
+    assert 'Traceback (most recent call last)' in result.stdout.str()
+    assert 'During handling of the above exception' not in result.stdout.str()
+    assert 'Differing items:' not in result.stdout.str()
```
The-Compiler: Hmm, maybe the test should make sure there is actually no second traceback (like reported in #44)?

nicoddemus: Good idea!

The-Compiler:
FWIW I don't really see a value in the warning. The monkeypatching is never something you'd enable explicitly, right? I'd imagine you either let it happen by default, or set `mock_traceback_monkeypatch = false` (Why does that option even exist? If something goes wrong with the patching?).

If I'm using `--tb=native` I'm already expecting longer/native/not-cleaned-up tracebacks, so it wouldn't surprise me at all if the monkeypatching doesn't take place. From the users perspective, if there wasn't #44, the patching wouldn't change *anything* in the output they see with `--tb=native`, right?

nicoddemus:
Note to self: mention this feature in `README`

nicoddemus:
> FWIW I don't really see a value in the warning. The monkeypatching is never something you'd enable explicitly, right?

Hmm you got a point, it is enabled by default so the user won't "try to use it and it won't work", which I agree would be confusing. @blueyed perhaps a note on the README is enough?

> Why does that option even exist? If something goes wrong with the patching?

I didn't know at the time if that monkeypatching could interfere with something else in more complex test suites, so I decided to put that option in there more like a preemptive measure.

blueyed:
> @blueyed perhaps a note on the README is enough?

Would be fine with me.
My thinking was that a user might wonder why it does not work, enables it explicitly even, and then it still does not work.  In the latter case a warning/note might be helpful, but could be easily overseen anyway.

nicoddemus:
Thanks guys!

PR #49, opened 2016-06-01
add 'mocker.call' convenience name (closes #48)

### LINKED ISSUE #48: Provide `mocker.call` name
In order to use the `assert_has_calls` check, you need the `call` name. I think it should be part of the set of 'convenience' names available via `mocker`.

The end goal is to make this work::

    get_pwd = mocker.patch('keyring.get_password')
    …
    get_pwd.assert_has_calls(
        mocker.call(Credentials.KEYRING_SERVICE_DEFAULT, 'jane@keyring.example.com'),
        mocker.call(Credentials.KEYRING_SERVICE_DEFAULT, 'keyring.example.com'))

If you agree it should be available, I can do a PR.

jhermann:
See #48 for motivation.

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -20,6 +20,7 @@ class MockFixture(object):
     Mock = mock_module.Mock
     MagicMock = mock_module.MagicMock
     PropertyMock = mock_module.PropertyMock
+    call = mock_module.call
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks! I will merge this later :grin:

PR #50, opened 2016-06-01
fix setuptools 'license' data to MIT

jhermann:
License switch now a :100:% complete :smirk_cat:


nicoddemus:
oops! thanks :grin:

PR #56, opened 2016-08-02
Added "mock.sentinel" to the "mocker" fixture for convenience.

kjwilcox:
Given that there is already some precedence for this sort of thing, it would be very convenient to be able to access `mock.sentinel` off of the `mocker` fixture.

I added this, updated the docs, updated the changelog, and added a test case (just adding a parameter to an existing parameterized test).

Please let me know whether or not this is a change you like, and if yes, if there's anything more I need to do to have this PR accepted.

Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@
   to install a newer ``mock`` version from PyPI instead of using the one available in the
   Python distribution.
   Thanks `@wcooley`_ for the PR (`#54`_).
+* ``mock.sentinel`` is now aliased as ``mocker.sentinel`` for convenience.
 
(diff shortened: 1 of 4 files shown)


nicoddemus:
That's perfect, thanks a lot! :grin:

kjwilcox:
Thanks for the super quick response!

Any idea when 1.2 might see a release?

nicoddemus:
Just pushed `1.2` to PyPI. :grin: :+1:

kjwilcox:
Excellent, thank you!

nicoddemus:
Thanks again for the PR! :+1:

Issue #61, opened 2016-09-03
Fix tests for pytest 3


vincentbernat:
Hey!

Many tests are failing when running with pytest 3.0.1:

```
_____________________________ TestMockerStub.test_failure_message_with_no_name _____________________________

self = <test_pytest_mock.TestMockerStub instance at 0x7f0b7f55c758>
mocker = <pytest_mock.MockFixture object at 0x7f0b7f51add0>

    def test_failure_message_with_no_name(self, mocker):
>       self.__test_failure_message(mocker)

test_pytest_mock.py:182:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_pytest_mock.TestMockerStub instance at 0x7f0b7f55c758>
mocker = <pytest_mock.MockFixture object at 0x7f0b7f51add0>, kwargs = {}, expected_name = 'mock'
expected_message = 'Expected call: mock()\nNot called'
stub = <MagicMock spec='function' id='139687357426384'>, exc_info = <ExceptionInfo AssertionError tblen=3>
@py_assert1 = AssertionError('Expected call: mock()\nNot called',)

    def __test_failure_message(self, mocker, **kwargs):
        expected_name = kwargs.get('name') or 'mock'
        expected_message = 'Expected call: {0}()\nNot called'.format(expected_name)
        stub = mocker.stub(**kwargs)
        with pytest.raises(AssertionError) as exc_info:
            stub.assert_called_with()
>       assert exc_info.value.msg == expected_message
E       AttributeError: 'exceptions.AssertionError' object has no attribute 'msg'

test_pytest_mock.py:179: AttributeError
```

blueyed:
Thanks for reporting.
I've rephrased the issue: it works with pytest 3 in general.
A PR would be welcome, of course.

vincentbernat:
I'll do a PR shortly.

nicoddemus:
Fixed in #63

PR #62, opened 2016-09-03
Update tests to work with pytest 3

vincentbernat:
Also update tox.ini to run them.

This is not complete as 2 tests are still failing. I don't grasp exactly why, but I suspect we get a regular `AssertionError()` because pytest didn't replace the `assert` at this point by its own one.

Related to #61.

### DIFF
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -176,7 +176,7 @@ class TestMockerStub:
         stub = mocker.stub(**kwargs)
         with pytest.raises(AssertionError) as exc_info:
             stub.assert_called_with()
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks! I'm surprised, the tests should be working.

I will take a closer look on monday.

nicoddemus:
Thanks for the PR. The tests are failing due to an unrelated issue related to how assertion reinterpretation was removed in pytest 3.0 and rewriting doesn't quite working automatically in development mode for plugins. 

I opened pytest-dev/pytest#1934 to discuss it.

PR #64, opened 2016-09-17
Fix for python 3.6a

### LINKED ISSUE #59: Error for Python 3.6 nightly
Hi,

our [cookiecutter build](https://travis-ci.org/audreyr/cookiecutter/jobs/149489032) breaks for Python 3.6 nightly due to some error in **pytest-mock**.

```
    class MockFixture(object):
TypeError: '_SentinelObject' object is not callable
```

Figured I'll raise this here although it's certainly not a problem at this point as Python 3.6 isn't out just yet.

nicoddemus:
Fix #59

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,11 @@
+1.3
+---
+
+* Add support for Python 3.6. Thanks `@hackebrot`_ for the report (`#59`_).
+
+.. _@hackebrot: https://github.com/hackebrot
+.. _#59: https://github.com/pytest-dev/pytest-mock/issues/59
+
 1.2
 ---
 
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -10,7 +10,7 @@ except ImportError:
 version = '1.2'
 
 
-class MockFixture(object):
+class MockFixture:
     """
     Fixture that provides the same interface to functions in the mock module,
     ensuring that they are uninstalled at the end of each test.
@@ -21,12 +21,15 @@ class MockFixture(object):
     PropertyMock = mock_module.PropertyMock
     call = mock_module.call
     ANY = mock_module.ANY
-    sentinel = mock_module.sentinel
+
 
     def __init__(self):
         self._patches = []  # list of mock._patch objects
         self._mocks = []  # list of MagicMock objects
         self.patch = self._Patcher(self._patches, self._mocks)
+        # temporary fix: this should be at class level, but is blowing
+        # up in Python 3.6
+        self.sentinel = mock_module.sentinel
 
     def resetall(self):
         """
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -1,3 +1,11 @@
+1.3
+---
+
+* Fix error when running ``pytest-mock`` in Python 3.6a. Thanks `@hackebrot`_ for the report (`#59`_).
```
blueyed: Also applies to 3.6.0b1, I suppose?!
Maybe just say 3.6 then?

nicoddemus: Good point, specially because it is now part of CI. Will change it.

hackebrot:
Nice one @nicoddemus @blueyed! 👍

pytest_mock.py
```diff
@@ -10,7 +10,7 @@
 version = '1.2'
 
 
-class MockFixture(object):
+class MockFixture:
```
The-Compiler: Wouldn't make this `MockFixture` an old-style class on Python 2? Is that intended?

nicoddemus: Oh right. I tried that first as a fix and forgot to revert it.

On the other hand being an old-style class doesn't really matter for `MockFixture` because it doesn't have any comparison operators or meta-classes, which is where old-style and new-style classes differ. Do you rather me revert that?

The-Compiler: I'd prefer that, as I have no idea how an old-style class works, and I'm sure I'm not the only one :wink:

blueyed: But apparently @nicoddemus knows how they work and that it does not matter in this case.
Makes the code cleaner, too - so I am +1 for keeping it like this.

blueyed: It is good btw to forget about old-style things.. ;)
We have tests for catching that hopefully.. :)

The-Compiler: Except that maybe @nicoddemus isn't the only one working on that code :wink:

And I agree about forgetting about old-style things, that's why it should be a new-style class :wink:

blueyed: @The-Compiler 
Well:

> % git shortlog -sn
   122  Bruno Oliveira
     4  Rafael Bertoldi (fogo)
     3  Jurko Gospodnetić
     2  Florian Bruhin
     2  Pavel Savchenko
     2  Tiago Nobrega
     2  jhermann

;)

But it's under the pytest-dev umbrella of course, so you are right in that regard.

new-style classes however are only relevant for py2, and I was disregarding this by ignoring old-style things.. ;)

nicoddemus: Since @The-Compiler prefers to bring it back, no worries! Pushed 931785c. 😁

PR #72, opened 2016-11-22
New mocker.mock_module variable points to the underlying mock module being used

### LINKED ISSUE #71: Alias _Call class?
I am/was using code like the following to detect if some object is a `_Call` instance (to stop recursion in this case):

```python
    from pytest_mock import mock_module

    try:
        _Call = mock_module.mock._Call
    except AttributeError:
        # unittest.mock
        _Call = mock_module._Call
    if isinstance(exp, _Call):
        ...
```

Since `mock_module` has been moved in https://github.com/pytest-dev/pytest-mock/commit/a4cfee7a0843a07e86a458a6c129893356c6b122, I wonder what is the best way to fix / address this.

nicoddemus:
Fix #71

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,18 @@
+1.5.0
+-----
+
+* New ``mocker.mock_module`` variable points to the underlying mock module being used
+  (``unittest.mock`` or ``mock``).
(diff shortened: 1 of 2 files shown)


nicoddemus:
`1.5.0` published. Thanks @blueyed! 😁

PR #77, opened 2017-02-15
remove duplicated assignment of self.sentinel and self.mock_open

kjwilcox:
I noticed this while I was browsing the source. Looks like a merge got screwed up at some point and these two lines got duplicated.


nicoddemus:
Oh thanks a lot for taking the time to write the PR, appreciate it!

PR #88, opened 2017-07-16
Only reset actual mocks in resetall().

jdavisp3:
The `mocker.patch.dict` method currently causes `mocker.resetall()` to fail, because `patch.dict` does not return a mock that can be reset.

* Only append the return value of the patching function to the list of outstanding mocks if it has the `reset_mock` attribute.
* Add a test verifying we can call `resetall()` after patching a dict.
* Add a test verifying we can call `resetall()` and `stopall()` after other patching methods.

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -124,7 +124,8 @@ class MockFixture(object):
             p = mock_func(*args, **kwargs)
             mocked = p.start()
             self._patches.append(p)
-            self._mocks.append(mocked)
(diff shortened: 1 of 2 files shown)


nicoddemus:
Awesome, thank you! I will publish a new release later. 👍

jdavisp3:
Thanks very much!

nicoddemus:
`1.6.2` has just been released @jdavisp3, thanks!

jdavisp3:
🚀

Issue #102, opened 2018-02-15
Test failures in test_assert_called_args_with_introspection


ghisvail:
Whilst attempting to integrate `pytest-mock` to the Debian CI infrastructure, I encountered the following issue when running the tests against the installed version from the Debian package:
```
=================================== FAILURES ===================================
__________________ test_assert_called_args_with_introspection __________________

mocker = <pytest_mock.MockFixture object at 0x7f46460fce10>

    def test_assert_called_args_with_introspection(mocker):
        stub = mocker.stub()
    
        complex_args = ('a', 1, set(['test']))
        wrong_args = ('b', 2, set(['jest']))
    
        stub(*complex_args)
        stub.assert_called_with(*complex_args)
        stub.assert_called_once_with(*complex_args)
    
        with assert_argument_introspection(complex_args, wrong_args):
            stub.assert_called_with(*wrong_args)
>           stub.assert_called_once_with(*wrong_args)

test_pytest_mock.py:390: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/contextlib.py:35: in __exit__
    self.gen.throw(type, value, traceback)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

left = ('a', 1, set(['test'])), right = ('b', 2, set(['jest']))

    @contextmanager
    def assert_argument_introspection(left, right):
        """
        Assert detailed argument introspection is used
        """
        try:
            yield
        except AssertionError as e:
            # this may be a bit too assuming, but seems nicer then hard-coding
            import _pytest.assertion.util as util
            # NOTE: we assert with either verbose or not, depending on how our own
            #       test was run by examining sys.argv
            verbose = any(a.startswith('-v') for a in sys.argv)
            expected = '\n  '.join(util._compare_eq_iterable(left, right, verbose))
>           assert expected in str(e)
[...]

PR #107, opened 2018-02-19
Fix Python 3 compatibility

ghisvail:
Otherwise, the build fails with an encoding error due to the default locale being ASCII on the Debian builders, not UTF-8. Regardless, this is the recommended approach for writing a `setup.py` which supports both Python 2 and Python 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 from setuptools import setup
 
+from io import open
 
 setup(
     name='pytest-mock',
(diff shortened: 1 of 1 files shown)


jurko-gospodnetic:
:+1:

nicoddemus:
Thanks @ghisvail, appreciate it. After we sort out #102 we can make a new release with this.

PR #108, opened 2018-03-01
Dont use EMOJI in README.

mimi1vx:
Fixes:
    /usr/bin/python3 setup.py build '--executable=/usr/bin/python3 -s'
         File "setup.py", line 23, in <module>
             long_description=open('README.rst').read(),
         File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
             return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 1820: ordinal not in range(128)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ of a test:
   :target: http://www.opensourcecitizen.org/project?url=github.com/pytest-dev/pytest-mock
 
 If you found this library useful, donate some CPU cycles to its
-development efforts by clicking above. Thank you! 😇  
+development efforts by clicking above. Thank you!
 
 Usage
 =====


blueyed:
I'd say that it's better to fix setup.py.

nicoddemus:
I agree with @blueyed, it is a matter of fixing `setup.py`:

```python
long_description=io.open('README.rst', encoding='UTF-8').read(),
```

nicoddemus:
Wait, this actually is fixed already:

https://github.com/pytest-dev/pytest-mock/blob/818f3f56c4ddf665702f24e201a7c69f204edd42/setup.py#L24

But the fix is not in `1.7.0`; I will make a new release.

nicoddemus:
@mimi1vx `1.7.1` is out. 👍

PR #111, opened 2018-04-06
Adds missing aliases for NonCallableMock and create_autospec

mlhamel:
This PR is adding support for the following missing aliases:
* `NonCallableMock` :: (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.NonCallableMock)
* `create_autospec` :: https://docs.python.org/3/library/unittest.mock.html#create-autospec

It simply adds the missing instantiation into `MockFixture` and adds it also into the corresponding test.


nicoddemus:
Thanks a lot @mlhamel!

nicoddemus:
`1.8.0` is out, thanks again!

PR #113, opened 2018-04-10
Add support for `assert_called_once()`

rouge8:
This was added in Python 3.6.

ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -221,6 +221,12 @@ def wrap_assert_called_with(*args, **kwargs):
                    *args, **kwargs)
 
 
+def wrap_assert_called_once(*args, **kwargs):
+    __tracebackhide__ = True
+    assert_wrapper(_mock_module_originals["assert_called_once"],
+                   *args, **kwargs)
+
+
 def wrap_assert_called_once_with(*args, **kwargs):
     __tracebackhide__ = True
     assert_wrapper(_mock_module_originals["assert_called_once_with"],
@@ -254,6 +260,7 @@ def wrap_assert_methods(config):
         'assert_not_called': wrap_assert_not_called,
         'assert_called_with': wrap_assert_called_with,
         'assert_called_once_with': wrap_assert_called_once_with,
+        'assert_called_once': wrap_assert_called_once,
         'assert_has_calls': wrap_assert_has_calls,
         'assert_any_call': wrap_assert_any_call,
     }
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -388,6 +388,17 @@ def test_assert_called_once_with_wrapper(mocker):
         stub.assert_called_once_with("foo")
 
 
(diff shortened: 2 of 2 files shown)


pytest_mock.py
```diff
@@ -254,6 +260,7 @@ def wrap_assert_methods(config):
         'assert_not_called': wrap_assert_not_called,
         'assert_called_with': wrap_assert_called_with,
         'assert_called_once_with': wrap_assert_called_once_with,
+        'assert_called_once': wrap_assert_called_once,
```
nicoddemus: You probably need to add this only in Python 3.6.

rouge8: It's definitely present in [Mock on PyPI](https://pypi.python.org/pypi/mock) too. I actually noticed it on Python 2.7 because `mocker.spy()` didn't have that method... Is there a nice way to tell if we're using the stdlib or external mock?

nicoddemus: Ahh good point. But I just noticed that what actually fails is the *test*, so you can change your test thus:

```python
    def test_assert_called_once_wrapper(mocker):
        stub = mocker.stub()
        if not hasattr(stub, 'assert_called_once'):
            pytest.skip('assert_called_once not available')
        stub("foo")
        ...
```

nicoddemus:
Thanks @rouge8!

rouge8:
@nicoddemus interestingly this isn't getting added to `mocker.spy()`. This test is skipping on Python 2:

```python
def test_assert_called_once_spy(mocker):
    class Foo(object):
        def bar(self):
            return 1
    f = Foo()
    mocker.spy(f, 'bar')
    if not hasattr(f.bar, 'assert_called_once'):
        pytest.skip('assert_called_once not available')

    f.bar()
    f.bar.assert_called_once()
```

But the test in this PR runs:

```python
def test_assert_called_once_wrapper(mocker):
    stub = mocker.stub()
    if not hasattr(stub, 'assert_called_once'):
        pytest.skip('assert_called_once not available')
    stub("foo")
    stub.assert_called_once()
    stub("foo")
    with assert_traceback():
        stub.assert_called_once()
```

nicoddemus:
@rouge8 yeah that seems weird. I can reproduce it with plain `mock` though:

```python
def test_assert_called_once_patch():
    import mock

    class Foo(object):
        def bar(self):
            return 1
    f = Foo()
    with mock.patch.object(f, 'bar', side_efect=f.bar, autospec=True):
        f.bar()
        f.bar.assert_called_once()
```

If I change `autospec` to `False`, then the test passes... 😕

rouge8:
Oh hey it's https://github.com/testing-cabal/mock/pull/420

nicoddemus:
Oh OK, then I guess we need to wait for a new `mock` release then. 😁

PR #115, opened 2018-04-10
Add support for `assert_called()`

rouge8:
This was added in Python 3.6.

I also re-ordered the wrappers to match the order in the unittest.mock docs to make it easier to see if any are missing in the future.

I'm not actually using this, but I noticed it while trying to figure out why `assert_called_once()` wasn't working with `mocker.spy()` 😂

ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+NEXT
+----
+
+* Add support for the recently added ``assert_called`` method in Python 3.6 and ``mock-2.0``. Thanks `@rouge8`_ for the PR (`#115`_).
+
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks a lot @rouge8, really appreciate you taking the time to write those PRs! 👍 

If you are not in a rush to use this, I will make a new release in a few weeks.

rouge8:
I don’t actually need this at all, so no rush!

nicoddemus:
OK thanks!

PR #120, opened 2018-08-12
Add `from pytest_mock import mocker` in the code example

Mariatta:


The-Compiler:
Thanks! However, I don't think that import should be needed, as `mocker` is used as a [pytest fixture](https://docs.pytest.org/en/latest/fixture.html) which don't rely on imports to work (for better or worse...).

nicoddemus:
Exactly. As @The-Compiler said, it is not really necessary to import `mocker` because it is a fixture which is injected by pytest automatically. Thanks for the PR though, appreciate it!

PR #138, opened 2019-03-25
Fix RuntimeError if user code calls patcher.stopall

### LINKED ISSUE #137: patch.stopall() breaks pytest-mock
I've just added `pytest` and `pytest-mock` to a legacy project and found that this error was being raised after tests completed.

```
  File "…/lib/python2.7/site-packages/pytest.py", line 89, in <module>
    raise SystemExit(pytest.main())
  File "…/lib/python2.7/site-packages/_pytest/config/__init__.py", line 80, in main
    return config.hook.pytest_cmdline_main(config=config)
  File "…/lib/python2.7/site-packages/pluggy/hooks.py", line 284, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "…/lib/python2.7/site-packages/pluggy/manager.py", line 68, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "…/lib/python2.7/site-packages/pluggy/manager.py", line 62, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "…/lib/python2.7/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "…/lib/python2.7/site-packages/pluggy/callers.py", line 81, in get_result
    _reraise(*ex)  # noqa
  File "…/lib/python2.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "…/lib/python2.7/site-packages/_pytest/main.py", line 236, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "…/lib/python2.7/site-packages/_pytest/main.py", line 231, in wrap_session
    config._ensure_unconfigure()
  File "…/lib/python2.7/site-packages/_pytest/config/__init__.py", line 648, in _ensure_unconfigure
    fin()
  File "…/lib/python2.7/site-packages/pytest_mock.py", line 291, in unwrap_assert_methods
    patcher.stop()
  File "…/lib/python2.7/site-packages/mock/mock.py", line 1513, in stop
    return self.__exit__()
  File "…/lib/python2.7/site-packages/mock/mock.py", line 1477, in __exit__
    raise RuntimeError('stop called on unstarted patcher')
RuntimeError: stop called on unstarted patcher
```

One of the tests in the test suite included a call to `patch.stopall()`, which seems to have stopped `pytest-mock`'s mocks and caused this error.

nicoddemus:
Fix #137

### DIFF
--- a/pytest_mock.py
+++ b/pytest_mock.py
@@ -288,7 +288,17 @@ def wrap_assert_methods(config):
 
 def unwrap_assert_methods():
     for patcher in _mock_module_patches:
-        patcher.stop()
+        try:
+            patcher.stop()
+        except RuntimeError as e:
+            # a patcher might have been stopped by user code (#137)
+            # so we need to catch this error here and ignore it;
+            # unfortunately there's no public API to check if a patch
+            # has been started, so catching the error it is
+            if text_type(e) == "stop called on unstarted patcher":
+                pass
+            else:
+                raise
     _mock_module_patches[:] = []
     _mock_module_originals.clear()
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,12 @@
+1.10.2
+------
+
+* Fix bug at the end of the test session when a call to ``patch.stopall`` is done explicitly by
+  user code. Thanks `@craiga`_ for the report (`#137`_).
+
+.. _#137: https://github.com/pytest-dev/pytest-mock/issues/137
+.. _@craiga: https://github.com/craiga
+
 1.10.1
 ------
 
--- a/test_pytest_mock.py
+++ b/test_pytest_mock.py
@@ -610,3 +610,24 @@ def test_assert_called_with_unicode_arguments(mocker):
 
     with pytest.raises(AssertionError):
         stub.assert_called_with(u"lak")
+
+
+def test_plain_stopall(testdir):
+    """Calling patch.stopall() in a test would cause an error during unconfigure (#137)"""
+    testdir.makepyfile(
+        """
+        import random
+
+        def get_random_number():
+            return random.randint(0, 100)
+
+        def test_get_random_number(mocker):
+            patcher = mocker.mock_module.patch("random.randint", lambda x, y: 5)
+            patcher.start()
+            assert get_random_number() == 5
+            mocker.mock_module.patch.stopall()
+    """
+    )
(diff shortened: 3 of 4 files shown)


blueyed:
Hmm, from https://github.com/pytest-dev/pytest-mock/issues/137#issuecomment-475669733 it sounds like it would be better if pytest-mock's mocks could be excluded?  But that's maybe a) not trivial, and b) not good in general - likely you should not call `stopall` if you do not mean it..

nicoddemus:
> likely you should not call stopall if you do not mean it..

Agree, unfortunately there's no public API to check for that; `__exit__` uses `_is_started(self)` to raise that exception, which I would use if it was public. I catching the exception and checking the actual message is safe, but brittle, as the message might change in future versions, so I think this is OK for now.

pytest_mock.py
```diff
@@ -288,7 +288,17 @@ def wrap_assert_methods(config):
...
+            patcher.stop()
+        except RuntimeError as e:
+            # a patcher might have been stopped by user code (#137)
+            # so we need to catch this error here and ignore it;
+            # unfortunately there's no public API to check if a patch
+            # has been started, so catching the error it is
+            if text_type(e) == "stop called on unstarted patcher":
+                pass
+            else:
+                raise
```
craiga: Thanks heaps for this, but I'm concerned that this might obfuscate other errors.

What would happen if some test called `.stopall()`, then some other test which relied on pytest-mock's internal mocks was called. Would it be clear what was going on?

craiga: This might be a separate issue. It's possible to get into this situation as-is.

nicoddemus: Thanks for the review!

I don't think so, as we are not catching any RuntimeError, we only ignore the very specific message... as I commented above, I think this is safe, but might break in future mock releases if they ever change the message. We should also switch to a public API that allows us to check if a patcher has already been stopped, if one is ever introduced.

tirkarthi:
Just to add to this the `RuntimeError` was removed in CPython and mock 3.0.0 due to similar concerns raised in #137 . Upstream issue : https://bugs.python.org/issue36366 .

nicoddemus:
Thanks @tirkarthi for the pointer, created #147 to track it. 👍

PR #141, opened 2019-03-30
tests: improve test_deprecated_mock

blueyed:
Use an inner test run to avoid the DeprecationWarning in the outer
tests.


nicoddemus:
Thanks!

PR #162, opened 2019-10-05
Do not show "pytest introspection" msg when there is no introspection

### LINKED ISSUE #154: pytest introspection with assert_called_once_with()
When a mock is called multiple times (with the right arguments, at some point) but only one call is expected:

```python
def test_foo(mocker):
    mock = mocker.Mock()
    mock('foo')
    mock('test')
    mock.assert_called_once_with('test')
```

the output is:

```
>       mock.assert_called_once_with('test')
E       AssertionError: Expected 'mock' to be called once. Called 2 times.
E       
E       pytest introspection follows:

test_x.py:5: AssertionError
```

which is kind of confusing.

nicoddemus:
Fix #154

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,14 @@
-1.11.1
-------
+1.11.2 (unreleased)
+-------------------
+
+* The *pytest introspection follows* message is no longer shown
+  if there is no pytest introspection (`#154`_).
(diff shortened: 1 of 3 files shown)


The-Compiler:
Looks good, thanks @nicoddemus! :+1:

Issue #164, opened 2019-10-21
mocker.patch.object appears to be broken


binarymason:
Python version: 3.7.3

Packages on my machine:

- pytest: 5.2.1
-  pytest-mock: 1.11.1

Testing a patched object method is called will fail using `pytest_mock`:

```python
import pytest
from pytest_mock import mocker

def test_process_is_saved_to_db(mocker):
    with mocker.patch.object(Process, 'save', return_value="it works") as mock_save:
        thing = Process()
        thing.save()

    mock_save.assert_called_once()
```

> AssertionError: Expected 'mock' to have been called once. Called 0 times.

<hr>

But will pass with the exact same signature using `unittest.mock`:

```python
import pytest
from unittest.mock import patch


def test_process_is_saved_to_db():
    with patch.object(Process, 'save', return_value="it works") as mock_save:
        thing = Process()
        thing.save()

    mock_save.assert_called_once()
```

> All tests passed

nicoddemus:
Hi @binarymason,

`mocker.patch.object` is not supposed to be used in a `with` context, please see [Note about usage as a contextmangager](https://github.com/pytest-dev/pytest-mock#note-about-usage-as-context-manager).

binarymason:
Ah, my bad. I need to RTFM.

Would it be helpful for future users if an exception were raised when `mocker.patch.object` is used in a `with` context?

Something along the lines of this?

```py
class MockFixture(object):
    def __enter__(self):
        raise ValueError("with contexts are not supported, etc...")

```

nicoddemus:
Hmmm that's a good idea. Would you like to open a PR in that effect?

PR #165, opened 2019-10-22
Raise an error if mocker is used in a `with` context

binarymason:
Ref: #164

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,11 +1,16 @@
-1.11.2 (unreleased)
+1.11.2 (2019-10-19)
 -------------------
 
 * The *pytest introspection follows* message is no longer shown
   if there is no pytest introspection (`#154`_).
   Thanks `@The-Compiler`_ for the report.
 
+* ``mocker`` now raises a ``TypeError`` when used as a context-manager.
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks a lot @binarymason!

1.11.2 released with this. 👍

svanburen:
fwiw, the CHANGELOG / release notes say `TypeError` while it looks like this actually raises a `ValueError`.

Also, I feel like this should probably catch situations in which multiple context managers are used?

```py
with context_manager_1, mocker.patch(...):
```

Thoughts? Might be harder to detect.

nicoddemus:
> fwiw, the CHANGELOG / release notes say TypeError while it looks like this actually raises a ValueError.

Oh right, I will update the CHANGELOG thanks

> Thoughts? Might be harder to detect.

I thought about that too, but I couldn't come up with a good way to detect that too. In the end though I think this is fine, this will probably catch 99% of the mistakes. 👍 

Thanks again!

liamdawson:
Feedback: whilst presumably for a good reason, this did introduce a breaking change for a rarely-built project of ours.

nicoddemus:
Thanks @liamdawson, can you share more details? This error was introduced because otherwise you would get a criptic message otherwise.

PR #173, opened 2019-11-27
Spy exceptions

inderpreet99:
On spied methods, we are capturing return value. However when exceptions are raised, `return_value` comes back as `False` rather than the exception itself.

This PR change stores the raised exception into the return value.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+1.13.0
+------
+
+* The object returned by ``mocker.spy`` now also tracks any side effect
+  of the spied method/function.
+
 1.12.1 (2019-11-20)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,5 @@
-1.13.0
-------
+1.13.0 (2019-12-05)
+-------------------
 
 * The object returned by ``mocker.spy`` now also tracks any side effect
   of the spied method/function.
--- a/README.rst
+++ b/README.rst
@@ -103,6 +103,9 @@ features with it, like retrieving call count. It also works for class and static
 Since version ``1.11``, it is also possible to query the ``return_value`` attribute
 to observe what the spied function/method returned.
 
+Since version ``1.13``, it is also possible to query the ``side_effect`` attribute
+to observe any exception thrown by the spied function/method.
+
 Stub
 ----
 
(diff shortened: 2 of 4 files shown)


nicoddemus:
Hi @inderpreet99, thanks for the PR.

When we want a mock to raise an exception, we usually set `side_effect`. Wouldn't be better to use `side_effect` then? Also, returning an exception value is a valid use case.

inderpreet99:
> Hi @inderpreet99, thanks for the PR.
> 
> When we want a mock to raise an exception, we usually set `side_effect`. Wouldn't be better to use `side_effect` then? Also, returning an exception value is a valid use case.

Hi @nicoddemus,

This PR is not addressing mocked methods raising exceptions, I see that `side_effect` can help with that nicely. This PR is for spying on methods that raise exceptions. They may go unnoticed to our test because maybe upstream in that code the exception would be silenced (for example) or it may be gobbled up into another `Exception`. This PR helps make an assertion on one of those spied methods (deep in the code).

I agree returning an exception is a valid use case, which this PR will not be able to distinguish. So this PR is probably not ideal. I've seen pytest code record `ExceptionInfo`s. Not sure how that stuff works.

Let me know if you have thoughts to solving the problem. I'm not tied to this approach.

nicoddemus:
Sorry I wasn't clear, but perhaps showing what I meant in code is simpler:

```python
        @w
        def wrapper(*args, **kwargs):
            try:
                r = method(*args, **kwargs)
            except Exception as e:
                result.side_effect = e
                raise
            else:
                result.return_value = r
            return r
```

inderpreet99:
@nicoddemus, thanks. I completely misread what you were saying. Just updated the code.

I had another problem with not being able to see multiple invocations of a spied method. For example, I think the above code will only store the latest return value.

What do you think about having return_values and side_effects as a list that maybe line up exactly with `mock_calls` or `call_args_list` for example?

Thanks.

nicoddemus:
I'm OK with us having both I guess, so keep the existing `return_value` and `side_effect`, but add `return_value_list` and `side_effect_list` variants? What do you think?

inderpreet99:
> Could you please also update the documentation showing the new feature?

Added

> I'm OK with us having both I guess, so keep the existing `return_value` and `side_effect`, but add `return_value_list` and `side_effect_list` variants? What do you think?

I'm facing a bug in regard to recording multiple side effects, so I'll get that code working in a separate branch. Especially since we are OK with keeping both features (last return/exception as well as all returns/exceptions).

nicoddemus:
1.13 released! Thanks a lot @inderpreet99!

PR #179, opened 2020-01-25
Merge pull request #179 from frankie567/issue-178

### LINKED ISSUE #178: Spy async bug introduced in Python 3.8.1
I believe [bpo-38875](https://bugs.python.org/issue38857) has introduced a bug in `mocker.spy()`. Spied coroutines are no longer awaited as they should be.

This test highlights the issue:

```python
import pytest


class MyClass:
    async def method(self):
        return "hi"


@pytest.mark.asyncio
async def test_spy_async(mocker):
    my_obj = MyClass()
    mocker.spy(my_obj, 'method')

    returned = await my_obj.method()

    # Passes
    assert my_obj.method.called

    # AssertionError: assert <coroutine object MyClass.method at 0x10f2a5d40> == 'hi'
    assert returned == "hi"
```

It looks like `spy()` (specifically, `wrapper()` defined within `spy()`) returns a coroutine and expects `unittest`'s `patch()` do to the awaiting, rather than doing the awaiting itself. This worked until bpo-38875 changed this behaviour. Now `unittest` will no longer await retuned awaitables, rather it returns them.

frankie567:
Hi there 👋

Took a shot at a fix for #178. Basically, we check if the spied function is a coroutine, and wrap it with an async-friendly wrapper.

Maybe not the most elegant solution, but I'm open to suggestions.

Cheers!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,10 +56,11 @@ jobs:
 
   deploy:
 
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+
     runs-on: ubuntu-latest
 
     needs: [build, linting]
-
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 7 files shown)


nicoddemus:
Excellent solution @frankie567, really appreciate it! Thanks!

PR #182, opened 2020-04-13
Add mocker for class, module, package and session scopes

### LINKED ISSUE #136: mocker fixture cannot be used in module scope
If i define a custom fixture like:

@pytest.fixture(scope='module')
def my_fixture(mocker):
    // code using mocker

class TestMyModule(object):

    def test_one(self, my_fixture):
        // test code

When i run this i get:
ScopeMismatch: You tried to access the 'function' scoped fixture 'mocker' with a 'module' scoped request object, involved factories

I'm i doing something wrong to do something like this? is there some reason why this is not supported?

Thanks.

scorphus:
While the feature proposed in pytest-dev/pytest#1681 is interesting and would serve as a base for a more elegant solution to #136, adding those supplementary mockers allows this plugin to be used in other scopes.

Fix #136

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -199,8 +199,7 @@ class MockFixture:
             return self._start_patch(self.mock_module.patch, *args, **kwargs)
 
 
-@pytest.yield_fixture
-def mocker(pytestconfig):
+def _mocker(pytestconfig):
     """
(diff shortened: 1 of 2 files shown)


scorphus:
My pleasure to contribute to this super useful package! Thanks for making it available, for taking the time to review and accept this PR and for drafting a new release! 💪

PR #187, opened 2020-05-07
Do not mark wheel as universal

### LINKED ISSUE #186: Wheel is incorrectly tagged as universal
``pytest-mock`` dropped Python 2 support in version 3.0.0 but wheel is still marked as universal (``py2.py3``) which is incorrect.

1. https://github.com/pytest-dev/pytest-mock/blob/81a35cfdb1d417674525c7cf015d77d570cf8f47/setup.cfg#L2
2. ``pytest_mock-3.1.0-py2.py3-none-any.whl`` at https://pypi.org/project/pytest-mock/#files

daa:
This pull-request removes marking ``pytest-mock`` wheel as universal because Python2 is no longer supported since version 3.0.0.

Fixes #186

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[bdist_wheel]
-universal=1


graingert:
the package already has a valid python_requires:

https://github.com/pytest-dev/pytest-mock/blob/master/setup.py#L11

nicoddemus:
> the package already has a valid python_requires:

Indeed, but we still shouldn't mark the wheel as universal I think. 😁

nicoddemus:
Thanks @daa!

daa:
Thank you!

PR #188, opened 2020-05-07
Revert "Do not mark wheel as universal"

graingert:
Reverts pytest-dev/pytest-mock#187

the package already has a valid python_requires:

https://github.com/pytest-dev/pytest-mock/blob/master/setup.py#L11
I think removing the py2 tag from the wheel will cause pip on py2 to download the sdist before seeing the python_requires

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal=1


The-Compiler:
What's the point of marking the wheel py2-compatible if it isn't? From https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels:

> Only use the `--universal` setting, if:
>
> 1. Your project runs on Python 2 and 3 with no changes (i.e. it does not require 2to3).
> 2. Your project does not have any C extensions.

and

> Pure Python Wheels that are not “universal” are wheels that are pure Python (i.e. contain no compiled extensions), but don’t natively support both Python 2 and 3.

graingert:
https://github.com/pypa/packaging.python.org/issues/726

PR #190, opened 2020-05-26
Permit warnings in *_stale_pyc tests

mgorny:
The *_stale_pyc test is repeatedly failing on production systems
due to additional pytest plugins being installed and triggering
PytestAssertRewriteWarnings.  This causes 'passed *' string not to match
because of the ',' after 'passed'.  Match just 'passed*' instead
to solve this.

### DIFF
--- a/tests/test_pytest_mock.py
+++ b/tests/test_pytest_mock.py
@@ -809,7 +809,7 @@ def test_abort_patch_context_manager_with_stale_pyc(testdir):
     """
     )
     result = testdir.runpytest()
-    result.stdout.fnmatch_lines("* 1 passed *")
+    result.stdout.fnmatch_lines("* 1 passed*")
 
     kwargs = {"legacy": True}
     assert compileall.compile_file(str(py_fn), **kwargs)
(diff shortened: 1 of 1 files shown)


nicoddemus:
Hi @mgorny,

Thanks for the patch!

> The *_stale_pyc test is repeatedly failing on production systems

Can you explain to me why are you running pytest-mock's tests in your production systems? 

The patch is simple of course, but I would like to know your use case better because it makes the tests a bit more fragile: besides warnings changing the message to `1 passed, 2 warnings in X seconds`, it might also hide other problems like errors and failures being given such as `1 passed, 1 error` in case of teardown failures.

mgorny:
We are running tests while installing to production systems to make sure that packages will actually work in the production system. As you have probably experienced yourself, there are some Python packages that break in the presence of other packages and testing them in pristine virtualenv doesn't guarantee that the same results will be reproducible in production.

nicoddemus:
Thanks @mgorny for context and sorry about the delay.

I've changed that test to use `result.assert_outcomes`, as that is safer.

nicoddemus:
Thanks!

mgorny:
Thank you a lot.

PR #194, opened 2020-06-25
Fix mypy errors

### LINKED ISSUE #193: Fix mypy errors
First step to provide type annotations support #152 (in order to do that, python 3.5 need to be dropped)

Current errors:
```
src/pytest_mock/plugin.py:5: error: Skipping analyzing 'pytest': found module but no type hints or library stubs
src/pytest_mock/plugin.py:5: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
src/pytest_mock/plugin.py:7: error: Skipping analyzing 'pytest_mock._version': found module but no type hints or library stubs
src/pytest_mock/plugin.py:210: error: Need type annotation for '_mock_module_originals' (hint: "_mock_module_originals: Dict[<type>, <type>] = ...")
src/pytest_mock/plugin.py:238: error: Assignment to variable 'e' outside except: block
Found 4 errors in 1 file (checked 1 source file)
```

staticdev:
Closes #193


staticdev:
@nicoddemus if you agree with this drop in 3.5, I can proceed with more strict typing support PRs.

nicoddemus:
Hi @staticdev,

First of all, thanks for tackling this!

> Notice this PR drops support for Python 3.5.

Can you explain why is that? I noticed you added an annotation to `_mock_module_originals`, is because of that? Can't we use a `#type:` comment instead?

staticdev:
@nicoddemus you are right. It is an alternative solution. This comment makes mypy ignore the type. By doing the way I did: `_mock_module_originals: Dict[str, Any]` mypy is actively resolving the types inside the script, so it. If you prefer I could change that to ignore.

nicoddemus:
>  If you prefer I could change that to ignore, but there is less benefits of type-hinting checks with that.

1. Drop Python 3.5 now and get full type-checking (merge as is and release now).
2. Use a type-hinting comment and get less benefits from type-checking (update the PR and release now).
3. Hold on to the PR until Python 3.5 is EOL to make the release.

The problem is that I wouldn't like to drop support for 3.5 just yet without a very strong reason, so I'm 👎 on 1).

What do you think?

staticdev:
> > If you prefer I could change that to ignore, but there is less benefits of type-hinting checks with that.
> 
>     1. Drop Python 3.5 now and get full type-checking (merge as is and release now).
> 
>     2. Use a type-hinting comment and get less benefits from type-checking (update the PR and release now).
> 
>     3. Hold on to the PR until Python 3.5 is EOL to make the release.
> 
> 
> The problem is that I wouldn't like to drop support for 3.5 just yet without a very strong reason, so I'm -1 on 1).
> 
> What do you think?

I think 3.5 EOL is just around the corner. You can see that [here](https://www.python.org/downloads/).

There is no problem of holding a release until then, but we could maybe create pre-releases and getting ready for that?

staticdev:
@nicoddemus @jtpavlock I managed to get this PR working WITH 3.5, based on pytest 6.0. 

This PR gives initial support to pytest-mock type-hinting. 

There are two catches:
* Pytest requires minimum of python 3.5.3 (you can see it [here](https://github.com/pytest-dev/pytest/blob/master/tox.ini) on Line 3).
* Variable annotations are not supported by 3.5 (as you can see [here](https://stackoverflow.com/questions/51789120/type-hints-syntax-error-on-python-3-5). So some refactoring may be needed to solve `_mock_module_originals` reverted [here](https://github.com/pytest-dev/pytest-mock/pull/194/commits/de9498c3088936f0fb575b30eeb4a01326fcc142).

nicoddemus:
> Pytest requires minimum of python 3.5.3 (you can see it here on Line 3).

That should be fine, as it should only affect users that want to use `mypy`. I think reasonable to have that requirement in that case.

> Variable annotations are not supported by 3.5 (as you can see here). So some refactoring may be needed to solve _mock_module_originals reverted here.

You can use type-hint comments for that:

```python
_mock_module_originals = {} # type: Dict[str, Any]
```

pytest uses it a lot because of lack of variable type annotations due to Python 3.5. I will re-add that to your PR. 👍

nicoddemus:
Thanks @staticdev!

PR #195, opened 2020-07-06
Fix link to docs.

tirkarthi:
Current link gives 404 and this seems to be the correct link as per pytest docs.


nicoddemus:
Thanks!

PR #197, opened 2020-07-07
Add pytest's introspection diff support for AsyncMock helpers

### LINKED ISSUE #196: Support introspection diffs for AsyncMock in Python 3.8
Python 3.8 added [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock) and added helpers equivalent to other assert helpers like assert_not_awaited, assert_awaited_with, etc. Currently diff support is added for Mocks. I guess it would be good to expand this to AsyncMock so that awaited calls also get better error messages. This would be similar to add to adding assert_wrapper for the helpers and patching the original helpers with this wrapped helpers as it's done now for. This is slightly related to #150 as it needs `AsyncMock` which is Python 3.8+ and mock 4.0+ to be available and thus the code might need conditionals to accommodate for the relevant versions.

A commit implementing this : https://github.com/tirkarthi/pytest-mock/commit/97b5800eb75360eec3cabf4a26698c787a3609bb

tirkarthi:
Fix #196

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,13 @@
+3.2.0 (2020-07-11)
+------------------
+
+* `AsyncMock <https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock>`__ is now exposed in ``mocker`` and supports provides assertion introspection similar to ``Mock`` objects.
+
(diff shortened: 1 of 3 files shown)


tirkarthi:
Thanks @nicoddemus for the merge. Thanks for all the work on pytest :)

PR #199, opened 2020-08-09
Stricter mypy

### LINKED ISSUE #152: add type annotations
I think it may be helpful to add type annotations to this package.

staticdev:
Closes #152


staticdev:
@nicoddemus I did a bunch of improvements over my initial PR #194

src/pytest_mock/plugin.py
```diff
@@ -72,9 +71,9 @@ def stopall(self):
         self._patches[:] = []
         self._mocks[:] = []
 
-    def spy(self, obj, name):
+    def spy(self, obj: Any, name: str) -> Any:
```
nicoddemus, line 74:
```suggestion
    def spy(self, obj: object, name: str) -> mock.MagicMock:
```

src/pytest_mock/plugin.py
```diff
@@ -134,7 +133,7 @@ async def async_wrapper(*args, **kwargs):
 
     def stub(self, name=None):
```
nicoddemus, line 134:
```suggestion
    def stub(self, name: str=None) -> mock.MagicMock:
```

staticdev:
@nicoddemus there is a problem with the changes you suggested. Since `mock` is imported via _get_mock_module and could be two different modules, it is not being resolved by type-check. Do you have any workaround for that? If not I suggest going back to Any or choose one of them to be imported in the beginning.

PS.: I have also added py.typed.

nicoddemus:
Hey @staticdev !

> If not I suggest going back to Any or choose one of them to be imported in the beginning.

Returning `Any` there will prevent users from benefit from type checking themselves, which defeats the point a bit. 

> Do you have any workaround for that? If not I suggest going back to Any or choose one of them to be imported in the beginning.

We can declare those functions as returning `unittest.mock.MagicMock` and use [typing.cast](https://docs.python.org/3/library/typing.html#typing.cast), for example:

```python
    def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
        ...
        spy_obj = self.patch.object(obj, name, side_effect=wrapped, autospec=autospec)
        spy_obj.spy_return = None
        spy_obj.spy_exception = None
        return cast(unittest.mock.MagicMock, spy_obj)
```

(We also need to import `unittest.mock` at the top level)

This works because both the standalone `mock` and `unittest.mock` implement the same interface, which is enough for the type checker.

Also we should document the other public functions with full arguments, for example `patch`, `multiple`, etc, like it was done in the [pytest_mock typeshed](https://github.com/python/typeshed/blob/master/third_party/3/pytest_mock/plugin.pyi), probably a few overloads too.

>  PS.: I have also added py.typed.

Thanks! We however also need to include that file in the package to signal the type checker that this module provides type annotations: https://www.python.org/dev/peps/pep-0561/#id18.

I understand this might seem quite a bit more work than it was thought initially, but I suspect it is needed to make the type annotations as useful as possible for users. 

I plan to tackle the issues above soon in case you don't have to work on them, probably in the next few days.

Thanks again all the work done so far.

staticdev:
Hi @nicoddemus,

About your last message:

1. I added the new changes to `spy()` you suggested, but with we have the errors:

```
src/pytest_mock/plugin.py:98: error: Too many arguments for "__getattribute__"
of "object"  [call-arg]
                    value = obj.__getattribute__(obj, name)
                            ^
src/pytest_mock/plugin.py:98: error: Argument 1 to "__getattribute__" of
"object" has incompatible type "object"; expected "str"  [arg-type]
                    value = obj.__getattribute__(obj, name)
                                                 ^
```

2. I tried documenting more methods using as a base https://github.com/python/typeshed/blob/master/third_party/3/pytest_mock/plugin.pyi. But, `_Patcher.object()` and `_Patcher.object(multiple)` have significantly different signatures and might need some refactors. Also, `parse_ini_boolean()` types are correct, but type-checks are still complaining about the code logic:

```
src/pytest_mock/plugin.py:411: error: Incompatible return value type (got
"Union[bool, str]", expected "bool")  [return-value]
            return value
                   ^
src/pytest_mock/plugin.py:413: error: Item "bool" of "Union[bool, str]" has no
attribute "lower"  [union-attr]
            return {"true": True, "false": False}[value.lower()]
                                                  ^
```

3. About py.typed https://www.python.org/dev/peps/pep-0561/#id18. I hope [this](https://github.com/pytest-dev/pytest-mock/pull/199/commits/216e7aa849a392f6ffcff404c6bd96be79190301) fixed that.

nicoddemus:
I see, thanks @staticdev!

I will take a look at the problems later then. 👍

PR #201, opened 2020-08-24
Type improvements

srittau:
* Annotate `MockerFixture.patch()`. Without an explicit annotations, mypy and other type checkers can't determine the proper type and assume `Any`.
* `MockerFixture.patch()` can return an arbitrary type if either `new` or `new_callable` is provided. Add overloads to handle these cases.

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -1,6 +1,6 @@
 import builtins
 import unittest.mock
-from typing import cast, Generator, Mapping, Iterable, Tuple
+from typing import cast, overload, Generator, Mapping, Iterable, Tuple, TypeVar
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -16,6 +16,8 @@ import functools
 import inspect
 
 import pytest
+
+_T = TypeVar("_T")
 
 
 def _get_mock_module(config):
@@ -50,7 +52,9 @@ class MockerFixture:
         self._patches = []  # type: List[Any]
         self._mocks = []  # type: List[Any]
         self.mock_module = mock_module = _get_mock_module(config)
-        self.patch = self._Patcher(self._patches, self._mocks, mock_module)
+        self.patch = self._Patcher(
+            self._patches, self._mocks, mock_module
+        )  # type: MockerFixture._Patcher
         # aliases for convenience
         self.Mock = mock_module.Mock
         self.MagicMock = mock_module.MagicMock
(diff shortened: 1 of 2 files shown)


src/pytest_mock/plugin.py
```diff
@@ -254,6 +258,63 @@ def dict(
...
+            self,
+            target: str,
+            new: None = ...,
+            spec: Optional[builtins.object] = ...,
+            create: bool = ...,
+            spec_set: Optional[builtins.object] = ...,
+            autospec: Optional[builtins.object] = ...,
+            new_callable: None = ...,
+            **kwargs: Any
+        ) -> unittest.mock.MagicMock:
```
graingert, line 272: these functions sometimes return ` unittest.mock.MagicMock:` and sometimes return `mock.MagicMock:`

nicoddemus: Indeed, but I don't think it matters to the type checker in the end, given that they both have the same interface, correct?

If not I don't know how to annotate that, given that the choice between mock/unittest.mock is done at runtime.

graingert: I think it might be possible with a


```python
class UnittestMockProtocol(Protocol): ...
class MockProtocol(Protocol): ...

_MP = TypeVar("_MP", bound=Union[UnittestMockProtocol, MockProtocol]])

class MockerFixture(Generic[_MP]): ...


MockMockerFixture = MockerFixture[MockProtocol]
UnittestMockerFixture = MockerFixture[UnittestMockProtocol]
```

and then write all the methods used twice over in each of UnittestMockProtocol and MockProtocol

graingert: otherwise codegen?

nicoddemus: Hmm interesting, but is it worth it? Don't `unittest.mock.MagicMock` and `mock.MagicMock` have the same interface?

nicoddemus: @graingert gentle ping.

graingert: they have a different interface when you're running a more modern backport, and they are nominal types

nicoddemus: Indeed that can happen, but I'm not sure this will happen often in practice vs doing something more complex like codegen or explicitly declaring protocols.

We can review this however in the future if I'm wrong on that, thanks for bringing it up. 👍

PR #203, opened 2020-08-24
remove redundant unittest.mock imports

graingert:


src/pytest_mock/plugin.py
```diff
@@ -87,7 +84,7 @@ def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
 
         :param object obj: An object.
         :param unicode name: A method in object.
-        :rtype: mock.MagicMock
+        :rtype: unittest.mock.MagicMock
```
graingert, line 87: do we care about `:rtype:`s here? can we use sphinx-autodoc-typehints instead?

nicoddemus: Hey @graingert! 

IMHO we can drop all types from the docs and rely solely on the type annotations.

About sphinx, we don't really have sphinx docs at the moment, so I don't think it matters.

Want to tackle this in this PR?

nicoddemus:
I don't think this applies anymore given we use `unittest.mock.MagicMock` for annotations.

Feel free to reopen if I'm wrong however!

graingert:
unittest.mock is imported 3 times, we only need it once

src/pytest_mock/plugin.py
graingert: I think worth defering to another pr

nicoddemus:
Ahh right, my bad! Sorry about that.

PR #205, opened 2020-08-25
Introduce MockFixture as alias to MockerFixture

### LINKED ISSUE #204: Major version increase?
Hi, 

I've seen 3.3.0 changed name from MockFixture to MockerFixture. It broke existing code like in:
https://github.com/staticdev/toml-validator/pull/157/checks?check_run_id=1020396590

I see at least two possible solutions here: increasing the major version to 4.0.0 to semantically say it broke original API or rename back (maybe even creating some alias to support both names).

nicoddemus:
Fix #204

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,15 @@
+3.3.1 (2020-08-24)
+------------------
+
+* Introduce ``MockFixture`` as an alias to ``MockerFixture``.
+
+  Before ``3.3.0``, the fixture class was named ``MockFixture``, but was renamed to ``MockerFixture`` to better
+  match the ``mocker`` fixture. While not officially part of the API, it was later discovered that this broke
+  the code of some users which already imported ``pytest_mock.MockFixture`` for type annotations, so we
+  decided to reintroduce the name as an alias.
+
+  Note however that this is just a stop gap measure, and new code should use ``MockerFixture`` for type annotations.
+
 3.3.0 (2020-08-21)
 ------------------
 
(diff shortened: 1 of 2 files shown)


nicoddemus:
`reveal_type` shows:

```
src\pytest_mock\__init__.py:4: note: Revealed type is 'def (config: Any) -> pytest_mock.plugin.MockerFixture'
```

So I believe that solves the issue?

graingert:
It looks like a mypy misconfiguration in both cases: missing [--disallow-any-unimported](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-unimported)

In 3.2.0 `pytest_mock.MockFixture` is an alias to Any, because it's not an exported symbol. But that also means that `pytest_mock.MockerFixture` is Any too!

so for all pytest_mock versions, with a misconfigured mypy:

```
if typing.TYPE_CHECKING:
    from pytest_mock import MockerFixture

def test_foo(mocker: "MockerFixture") -> None:
    pass
```

Will work for all pytest_mock versions


I think we should push back on this, and people using un-exported symbols in general because they are accidentality introducing Any

nicoddemus:
That's a good point, it wasn't doing any good before anyway.

Let's hold into this PR and see if more people break because of this to warrant adding a workaround.

The-Compiler:
Calling the default configuration a misconfiguration is... quite a stretch. Also note there are people using type annotations for other purposes (IDE autocompletion, human readers) which might work differently.

nicoddemus:
Hmm you are right, personally I meant more in the sense of it being `Any`, but I did not consider other uses other than mypy (such as IDEs).

@The-Compiler what's your opinion? Should we merge and release this, or just move on?

nicoddemus:
Yet another user bitten by this: https://github.com/pytest-dev/pytest-mock/pull/206

Let's go ahead and just get this in then, the maintenance cost is nearly zero.

graingert:
@The-Compiler this would fail with the default configuration, mypy doesn't allow untyped imports

PR #208, opened 2020-08-29
remove type annotations from docstrings

graingert:


src/pytest_mock/plugin.py
```diff
@@ -86,9 +86,8 @@ def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
         Create a spy of method. It will run method normally, but it is now
         possible to use `mock` call features with it, like call count.
 
-        :param object obj: An object.
-        :param unicode name: A method in object.
-        :rtype: unittest.mock.MagicMock
+        :param obj: An object.
```
graingert, line 89: do we need this either?

image

nicoddemus: Probably not hehehe

PR #209, opened 2020-09-10
Set a new fixture type

Abstract-X:
This PR replaces the deprecated fixture type `yield_fixture` with `fixture`.
> .. deprecated:: 3.0
Use :py:func:`pytest.fixture` directly instead.

[Link to docstring](https://github.com/pytest-dev/pytest/blob/9c0e0c756a36badb84edc017a803b9b0c4c122da/src/_pytest/fixtures.py#L1326-L1327)


nicoddemus (approved):
Thanks, appreciate it!

PR #214, opened 2020-11-10
Add resetall() arguments

### LINKED ISSUE #213: Add reset_mock()'s argument's to resetall()
`unittest`'s `Mock.reset_mock()` method has two keyword-only arguments `return_value` and `side_effect`, allowing to reset these attributes on mocks if desired ([unittest doc](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.reset_mock)). However `pytest-mock`'s `mocker.resetall()` doesn't have these arguments, requiring to reset mocks individually and breaking `resetall()`'s usefulness in that case.

The `return_value` and `side_effect` keyword-only arguments could be added to `resetall()` and passed to `reset_mock()` calls, mimicking its signature and behavior.

alex-marty:
Adding standard `unittest.reset_mock()`'s arguments to `MockerFixture.resetall()`. It seems like a small fix, please tell me if I'm missing something!

Closes #213

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -68,8 +68,13 @@ class MockerFixture:
         if hasattr(mock_module, "seal"):
             self.seal = mock_module.seal
 
-    def resetall(self) -> None:
-        """Call reset_mock() on all patchers started by this fixture."""
+    def resetall(self, *, return_value: bool = False, side_effect: bool = False) -> None:
+        """
+        Call reset_mock() on all patchers started by this fixture.
+        
+        :param bool return_value: Reset the return_value of mocks.
+        :param bool side_effect: Reset the side_effect of mocks.
+        """
         for m in self._mocks:
             m.reset_mock()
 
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks a lot for the PR, appreciate it!

The two keyword only arguments were added in Python 3.6 however, so it is breaking the 3.5 builds. I see two options:

1. Hold off the PR until we drop Python 3.5 support.
2. Implement the arguments conditionally, making them available only in Python 3.6+.

Which one do you prefer? 

I don't see many reasons to keep supporting Python 3.5 given it has reached EOL and pytest 6.2 won't support it anymore either, but it is not something I plan to tackle in the next few weeks.

alex-marty:
Hi! Tanks for the feedback! Indeed I didn't notice the issue with Python 3.5. I usually don't like implementing functionality conditional on runtime versions unless absolutely necessary, so my first impulse would be to wait until support of 3.5 is dropped, especially if it's just around the corner. There are easy workarounds for this functionality in the meantime. What do you think?

nicoddemus:
I also prefer to avoid them if possible.

So let's wait then. I will update the title to reflect this, and then I will pick it up once support for 3.5 is dropped. 👍 

Thanks again!

nicoddemus:
Time to drop Python 3.5. 🎉

nicoddemus:
Thanks again for the PR!

PR #221, opened 2020-12-22
Warn instead of raising exception in context manager

iforapsy:
Instead of raising an exception, warn when a pytest-mock mock is used as
a context manager. This lets pytest-mock mock objects that are used as
context managers, like threading.Lock, while still letting users know
that mocks returned from pytest-mock do not need their __enter__ method
called to take effect.

Create a PytestMockWarning class so that the warning is easy to
pinpoint and handle.

#192

### DIFF
--- a/src/pytest_mock/__init__.py
+++ b/src/pytest_mock/__init__.py
@@ -5,6 +5,7 @@ MockFixture = MockerFixture  # backward-compatibility only (#204)
 __all__ = [
     "MockerFixture",
     "MockFixture",
+    "PytestMockWarning",
     "pytest_addoption",
     "pytest_configure",
     "session_mocker",
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -234,14 +234,14 @@ class MockerFixture:
 
         def context_manager(
             self,
-            target: object,
+            target: builtins.object,
             attribute: str,
-            new: object = DEFAULT,
-            spec: Optional[object] = None,
+            new: builtins.object = DEFAULT,
+            spec: Optional[builtins.object] = None,
             create: bool = False,
-            spec_set: Optional[object] = None,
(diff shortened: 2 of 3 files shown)


nicoddemus:
Hey @iforapsy,

Thanks a lot for the PR, sorry for the delay!

The patch looks great, it shows you took great care on it, thanks.

One suggestion is to perhaps tweak the warning message a bit explaining more what's happening? Otherwise when using it as a context, the user might the impression that something is not correct.

---

Alternatively (and this just occurred to me now), perhaps we can have a different function, which operates like `mocker.patch.object`, but specific to test context managers and which doesn't emit a warning? Something like:

```python

import threading

class Foo:

    CREATION_LOCK = threading.Lock()
    def create_widgets(self):
        with self.CREATION_LOCK:
            print('Creating object')


def test_create_widgets_acquires_lock(mocker):
    """Test that the creation lock is acquired."""
    lock_mock = mocker.patch.context_manager(Foo, "CREATION_LOCK", autospec=True)
    Foo().create_widgets()
    lock_mock.__enter__.assert_called()
```

What do you think?

iforapsy:
Hi @nicoddemus,

No need to apologize. I understand that it's the holidays. I have not been the fastest responder either.

Thank you for your kind feedback. I like both your suggestions.

I made the warning message more explanatory and more newbie-friendly.

Though it won't be useful for my code base (as the `CREATION_LOCK` object is global in mine), I like the `mocker.patch.context_manager` function idea. It could be useful to others so I added that feature.

nicoddemus:
Hi @iforapsy, just passing by to let you know I did not forget about this PR, I will probably have time to review/merge/make a new release tomorrow. Thanks again for the patience!

nicoddemus (approved):
Great PR, thanks @iforapsy!

nicoddemus:
`3.5.0` released 🎉

PR #224, opened 2021-01-07
use getattr_static in spy instead of __getattributes__

yesthesoup:
When attempting to use `mocker.spy` on a Stripe class in my project, I ran into this error:

```python
    def test_create_customer(self, mocker):
>       stripe_spy = mocker.spy(stripe.Customer, 'create')

tests/[...]/test_stripe_customer_service.py:16:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pytest_mock.plugin.MockerFixture object at 0x10d6ebdd0>, obj = <class 'stripe.api_resources.customer.Customer'>, name = 'create'

    def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
        """
        Create a spy of method. It will run method normally, but it is now
        possible to use `mock` call features with it, like call count.

        :param obj: An object.
        :param name: A method in object.
        :return: Spy object.
        """
        method = getattr(obj, name)

        autospec = inspect.ismethod(method) or inspect.isfunction(method)
        # Can't use autospec classmethod or staticmethod objects
        # see: https://bugs.python.org/issue23078
        if inspect.isclass(obj):
            # Bypass class descriptor:
            # http://stackoverflow.com/questions/14187973/python3-check-if-method-is-static
            try:
>               value = obj.__getattribute__(obj, name)  # type:ignore
E               TypeError: descriptor '__getattribute__' requires a 'dict' object but received a 'type'

venv/lib/python3.7/site-packages/pytest_mock/plugin.py:110: TypeError
```

[Here is the Customer source code for reference](https://github.com/stripe/stripe-python/blob/master/stripe/api_resources/customer.py), a class inheriting from multiple parent classes and including some custom method wrappers.

While digging in order to fix this, I ended up at [the stackoverflow question linked to in the original code](https://stackoverflow.com/questions/14187973/python3-check-if-method-is-static) which mentioned the `inspect` module's method `getattr_static` available as of 3.2. This will accomplish the same behaviour as before, as well as work for a larger variety of classes.

I also checked up on the bug that this workaround is there for in the first place, and it was fixed in 3.7 it seems, so eventually autospec can be true by default in theory.

I don't quite know how to reproduce the exact kind of class that would have caused this error in the first place, open to suggestions in order to add a test for it.

### DIFF
--- a/tests/test_pytest_mock.py
+++ b/tests/test_pytest_mock.py
@@ -7,7 +7,7 @@ from typing import Callable, Any, Tuple, Generator, Type
 from unittest.mock import MagicMock
 
 import pytest
-from pytest_mock import MockerFixture, PytestMockWarning
+from pytest_mock import MockerFixture, PytestMockWarning  # type: ignore
 
 pytest_plugins = "pytester"
 
@@ -162,7 +162,7 @@ def test_mock_patch_dict_resetall(mocker: MockerFixture) -> None:
     ],
 )
 def test_mocker_aliases(name: str, pytestconfig: Any) -> None:
-    from pytest_mock.plugin import _get_mock_module
+    from pytest_mock.plugin import _get_mock_module  # type: ignore
 
     mock_module = _get_mock_module(pytestconfig)
 
@@ -268,6 +268,19 @@ def test_instance_method_spy_exception(
         assert str(spy.spy_exception) == "Error with {}".format(v)
 
 
+def test_instance_method_spy_autospec_true(mocker: MockerFixture) -> None:
+    class Foo:
+        def bar(self, arg):
+            return arg * 2
+
+    foo = Foo()
+    spy = mocker.spy(foo, "bar")
+    with pytest.raises(
+        AttributeError, match="'function' object has no attribute 'fake_assert_method'"
+    ):
+        spy.fake_assert_method(arg=5)
+
+
 def test_spy_reset(mocker: MockerFixture) -> None:
     class Foo(object):
         def bar(self, x):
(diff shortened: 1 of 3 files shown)


tests/test_pytest_mock.py
```diff
@@ -162,7 +162,7 @@ def test_mock_patch_dict_resetall(mocker: MockerFixture) -> None:
     ],
 )
 def test_mocker_aliases(name: str, pytestconfig: Any) -> None:
-    from pytest_mock.plugin import _get_mock_module
+    from pytest_mock.plugin import _get_mock_module  # type: ignore
```
yesthesoup, line 165: the pre-commit mypy run caught these imports with `error: Cannot find implementation or library`
but now the linting test is failing because of them...any ideas?

nicoddemus: Not sure, but I've moved `_get_mock_module` and `parse_ini_boolean` to a `_util.py` module: this makes it clear this is not part of the API plus should fix that particular problem.

nicoddemus:
Awesome, thanks!

nicoddemus:
`3.5.1` has been released. 🎉 

Thanks again!

yesthesoup:
> `3.5.1` has been released. 🎉
> 
> Thanks again!

Thanks for your help on this one!

PR #233, opened 2021-03-19
Fixing link to pytest assertions

fhightower:


nicoddemus:
Thanks! Failure is unrelated (probably time to drop 3.5).

PR #235, opened 2021-04-19
Correct type annotations in patch.dict()

plannigan:
[The documentation for `patch.dict()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict) says:
> `in_dict` can be a dictionary or a mapping like container. If it is a mapping then it must at least support getting, setting and deleting items plus iterating over keys.
> `in_dict` can also be a string specifying the name of the dictionary, which will then be fetched by importing it.

However, the argument is currently annotation is `Mapping[Any, Any]`. This updates the annotation to also include the string option.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,12 @@
+3.5.2 (UNRELEASED)
+------------------
+
+* Correct type annotations for ``mocker.patch.object`` to also include the string form.
+  Thanks `@plannigan`_ for the PR (`#235`_).
+
+.. _@plannigan: https://github.com/plannigan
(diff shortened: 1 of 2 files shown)


plannigan:
It looks like the CI failures are not related to my changes.
* `setuptools_scm` recently released v6.0, which dropped support for python 3.5
* The linting stage is setup to use python 3.7, but `tox` is trying to use python 3.6.

nicoddemus:
Thanks, looks good!

Could you also add a CHANGELOG entry?

plannigan:
Sure, updated now.

nicoddemus:
THanks a lot!

PR #241, opened 2021-05-05
modify resetall to work with patch object

### LINKED ISSUE #237: mocker.resetall() doesn't work anymore when using mocker.spy()
Using the same example from the README about `.spy()` but adding a `mocker.resetall()`
```
$ cat test_foo.py 
def test_spy_method(mocker):
    class Foo(object):
        def bar(self, v):
            return v * 2

    foo = Foo()
    spy = mocker.spy(foo, 'bar')
    assert foo.bar(21) == 42

    spy.assert_called_once_with(21)
    assert spy.spy_return == 42

    mocker.resetall()

$ pytest test_foo.py 
============================================= test session starts ==============================================
platform linux -- Python 3.8.0, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
rootdir: /home/blaxter
plugins: mock-3.6.0
collected 1 item                                                                                               

test_foo.py F                                                                                            [100%]

=================================================== FAILURES ===================================================
_______________________________________________ test_spy_method ________________________________________________

mocker = <pytest_mock.plugin.MockerFixture object at 0x7f376c49ddc0>

    def test_spy_method(mocker):
        class Foo(object):
            def bar(self, v):
                return v * 2
    
        foo = Foo()
        spy = mocker.spy(foo, 'bar')
        assert foo.bar(21) == 42
    
        spy.assert_called_once_with(21)
        assert spy.spy_return == 42
    
>       mocker.resetall()

test_foo.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pytest_mock.plugin.MockerFixture object at 0x7f376c49ddc0>

    def resetall(
        self, *, return_value: bool = False, side_effect: bool = False
    ) -> None:
        """
        Call reset_mock() on all patchers started by this fixture.
    
        :param bool return_value: Reset the return_value of mocks.
        :param bool side_effect: Reset the side_effect of mocks.
        """
        for m in self._mocks:
>           m.reset_mock(return_value=return_value, side_effect=side_effect)
E           TypeError: reset_mock() got an unexpected keyword argument 'return_value'

.virtualenvs/test/lib/python3.8/site-packages/pytest_mock/plugin.py:73: TypeError
=========================================== short test summary info ============================================
FAILED test_foo.py::test_spy_method - TypeError: reset_mock() got an unexpected keyword argument 'return_value'
============================================== 1 failed in 0.02s ===============================================
 ```

### PR DESCRIPTION
**Pull request by shadycuz**
Fixes #237

It seems `spy` and `patch.object` have a `reset_mock` that doesn't take the `return_value` and `side_effect` keyword args.

I originally tried to fix this in `_start_patch`
https://github.com/pytest-dev/pytest-mock/blob/c3878181c9ed65d748694e0f4b1eb58bef3dfce5/src/pytest_mock/plugin.py#L171-L178

but changing `self._mocks.append(mocked) ` to only append actually `Mock` objects breaks my repo worse because then I lose the ability to call `resetall` on patched objects.

This seemed like the best approach to take. 

I also tried to get this to fail in a test case but could not get the error to trigger. I suspect the issue might be with the version of `mock` or `unittest.mock` that our projects are using. I should be on the lastest of everything, but I'm still facing the error which is strange 🤔 

Here is my repo https://github.com/DontShaveTheYak/cloud-radar/pull/32 if you want to try and debug it.

### DIFF
--- a/tests/test_pytest_mock.py
+++ b/tests/test_pytest_mock.py
@@ -310,7 +310,7 @@ def test_spy_reset(mocker: MockerFixture) -> None:
     assert spy.spy_return == 30
     assert spy.spy_exception is None
 
-    # Testing spy can still be reset.
+    # Testing spy can still be reset (#237).
     mocker.resetall()
 
     with pytest.raises(ValueError):
--- a/tests/test_pytest_mock.py
+++ b/tests/test_pytest_mock.py
@@ -310,6 +310,9 @@ def test_spy_reset(mocker: MockerFixture) -> None:
     assert spy.spy_return == 30
     assert spy.spy_exception is None
 
+    # Testing spy can still be reset.
+    mocker.resetall()
+
     with pytest.raises(ValueError):
         Foo().bar(0)
     assert spy.spy_return is None
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,12 @@
+3.6.1 (UNRELEASED)
+------------------
+
+* Fix ``mocker.resetall()`` when using ``mocker.spy()`` (`#237`_). Thanks `@blaxter`_ for the report and `@shadycuz`_ for the PR.
+
+.. _@blaxter: https://github.com/blaxter
+.. _@shadycuz: https://github.com/shadycuz
+.. _#237: https://github.com/pytest-dev/pytest-mock/issues/237
+
 3.6.0 (2021-04-24)
 ------------------
 
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -16,6 +16,7 @@ from typing import Mapping
 from typing import Optional
 from typing import overload
 from typing import Tuple
+from typing import Type
 from typing import TypeVar
 from typing import Union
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: tests/test_pytest_mock.py [hunk 1]
```diff
@@ -310,6 +310,9 @@ def bar(self, x):
     assert spy.spy_return == 30
     assert spy.spy_exception is None
 
+    # Testing spy can still be reset.
+    mocker.resetall()
+
```
shadycuz, line 315: Found a working test case. If you add this line to the previous version `3.6.0` then it triggers the error.

shadycuz:
I now believe the failure happens when `autospec = True` but I don't know why 🤷‍♂️. That's the only thing in common between my failure with `patch.object` and the failure with `spy`. I think they both are using `autospec`. I read through the `mock` [source code](https://github.com/python/cpython/blob/3.9/Lib/unittest/mock.py) but couldn't find the reason.

nicoddemus:
> I now believe the failure happens when autospec = True but I don't know why 🤷‍♂️.

Yeah autospec is a bit convoluted... IIRC we can't use autospec with classmethod (or staticmethod) for example.

nicoddemus:
Thanks a lot @shadycuz, appreciate it!

PR #296, opened 2022-06-19
Add `async_stub` method

PerchunPak:
Which actually the same as `stub` but creates async stub.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -110,3 +110,7 @@ It may receive an optional name that is shown in its ``repr``, useful for debugg
 
         foo(stub)
         stub.assert_called_once_with('foo', 'bar')
+
+.. seealso::
+
+    ``async_stub`` method, which actually the same as ``stub`` but makes async stub.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@ Releases
 ========
 
 
+3.8.0 (unreleased)
+------------------
+* Add ``MockerFixture.async_mock`` method. Thanks `@PerchunPak`_ for the PR.
+
+.. _@PerchunPak: https://github.com/PerchunPak
+
 3.7.0 (2022-01-28)
 ------------------
 
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -159,6 +159,19 @@ class MockerFixture:
             self.mock_module.MagicMock(spec=lambda *args, **kwargs: None, name=name),
         )
 
+    def async_stub(self, name: Optional[str] = None) -> unittest.mock.AsyncMock:
+        """
+        Create a async stub method. It accepts any arguments. Ideal to register to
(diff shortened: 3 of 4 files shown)


docs/usage.rst
```diff
@@ -110,3 +110,7 @@ It may receive an optional name that is shown in its ``repr``, useful for debugg
 
         foo(stub)
         stub.assert_called_once_with('foo', 'bar')
+
+.. seealso::
+
+    ``async_stub`` method, which actually the same as ``stub`` but makes async stub.
```
PerchunPak, line 116: image

nicoddemus (changes requested):
Thanks a lot @PerchunPak for the contribution, sorry about the delay. 

There are some things that need fixing:

1. Can you take a look at the test failures? Seems like we have to put a guard around the definition of `async_stub` on the Python versions that don't support it.
2. Could you add a very simple test to `TestMockerStub` that just calls `async_stub` and ensures it is an instance of `unittest.mock.AsyncMock`? This is just to avoid future regressions. The test needs to be skipped on unsupported Python versions.
3. Finally could add a quick changelog entry for this, under a new session `3.8.0 (unreleased)`?

Thanks again!

PerchunPak:
> Can you take a look at the test failures? Seems like we have to put a guard around the definition of async_stub on the Python versions that don't support it.

Can't imagine how, seems like `mock_module.AsyncMock` will not work with type checkers, so how this could be done?
Maybe import standalone `mock` module in `if TYPE_CHECKING:` block?

nicoddemus:
Sorry I should have mention, but [mypy supports `sys.version_info` checks](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks). 👍

PerchunPak:
> Sorry I should have mentioned, but [mypy supports `sys.version_info` checks](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks). 👍

Yes, but how about runtime? Yes `unittest.mock` doesn't have `AsyncMock` in py3.7, but `mock` have.

nicoddemus:
I'm OK with only supporting AsyncMock for Python 3.8+, or is this something you need in Python 3.7?

PerchunPak:
> I'm OK with only supporting AsyncMock for Python 3.8+, or is this something you need in Python 3.7?

No, just you support both (`unittest.mock` and `mock`), it's strange that you breaking this support here.

nicoddemus:
Hmm an alternative would be then:

```python
if sys.version_info[:2] >= (3, 8):
    AsyncMockType = unittest.mock.AsyncMock
else:
    AsyncMockType = mock.AsyncMock
```

at the top of the file, then use `AsyncMockType` in the annotation. If the above does not work, then `AsyncMockType = Any` for Python 3.7 is a good compromise.

nicoddemus:
Awesome, thanks a lot for the contribution and patience! 🙇

PR #306, opened 2022-07-08
Remove unnecessary import

cclauss:
This import was useful for Python 2 but is no longer needed in Python 3.
* https://github.com/asottile/pyupgrade#open-alias


nicoddemus (approved):
Thanks!

PR #308, opened 2022-07-15
Add docs/_static directory to allow sphinx-build to work cleanly

juliangilbey:
The `sphinx-build` command expects to find a `docs/_static` directory, as specified in `docs/conf.py`.  This patch creates an essentially-empty directory for this purpose.  (I learnt this trick from https://github.com/chrislit/abydos)


nicoddemus:
Thanks

PR #319, opened 2022-10-01
Implement selective un-spying and un-patching

### LINKED ISSUE #259: Guidance on "un-spying" a method
Hi, I couldn't quite figure out how to un-spy a spied method yet. I'm not sure if it's even possible after browsing through the source code. 

```python
spy_foo = mocker.spy(module, "foo")

# do something

# how to un-spy `spy_foo`?
```

Any ideas or suggestions on how to achieve it? 🙏

sgaist:
This merge request implements the possibility to stop specific spies rather than having to just stop them all.

Based on the patch of @nicoddemus in issue #259

Fixes #259

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -21,6 +21,7 @@ The supported methods are:
 * `mocker.patch.multiple <https://docs.python.org/3/library/unittest.mock.html#patch-multiple>`_
 * `mocker.patch.dict <https://docs.python.org/3/library/unittest.mock.html#patch-dict>`_
 * `mocker.stopall <https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.stopall>`_
+* `mocker.stop <https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop>`_
 * ``mocker.resetall()``: calls `reset_mock() <https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.reset_mock>`_ in all mocked objects up to this point.
 
 Also, as a convenience, these names from the ``mock`` module are accessible directly from ``mocker``:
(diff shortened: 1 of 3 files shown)


src/pytest_mock/plugin.py
```diff
@@ -94,10 +95,16 @@ def stopall(self) -> None:
...
-        self._patches[:] = []
-        self._mocks[:] = []
+        self._patches_and_mocks.clear()
+
+    def stop(self, mock: unittest.mock.MagicMock) -> None:
+        for index, (p, m) in enumerate(self._patches_and_mocks):
+            if mock is m:
+                p.stop()
+                del self._patches_and_mocks[index]
+                break
```
nicoddemus, line 107: Probably it is best to raise an error if the given mock is not registered by us, to avoid mistakes being silent.

docs/usage.rst
```diff
@@ -94,6 +95,24 @@ As of version 3.0.0, ``mocker.spy`` also works with ``async def`` functions.
 
     .. _#175: https://github.com/pytest-dev/pytest-mock/issues/175
 
+As of version 3.9.1, spying can be also selectively stopped.
```
nicoddemus, line 98: Given this is a new feature, the next will be a minor release.

```suggestion
As of version 3.10, spying can be also selectively stopped.
```

nicoddemus (changes requested):
Thanks @sgaist! 

Looks good, left two minor comments.

src/pytest_mock/plugin.py
sgaist: Good point

docs/usage.rst
sgaist: Indeed, 3.10 or rather 3.10.0 ?

nicoddemus:
Thanks @sgaist!

PR #320, opened 2022-10-02
Add Python 3.11 support

gruebel:
pretty straight forward, no issues and deprecation warnings

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,8 @@ jobs:
             tox_env: "py39"
           - python: "3.10"
             tox_env: "py310"
+          - python: "3.11-dev"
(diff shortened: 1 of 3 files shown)


nicoddemus:
Strange:

image

Doesn't seem related to the PR, will take a look later.

gruebel:
@nicoddemus I think it was my mistake, I g=somehow forgot to add the Python version `3.11-dev` also to the `python` list in the matrix 🙈

nicoddemus:
Ahh right! Did not catch that because the error was mentioning `runs-on` and ` ${{ matrix.os }}`. 🤦‍♂️

nicoddemus:
Thanks!

PR #337, opened 2022-12-16
Remove old tox env (linting) from contributing doc

attomos:
Since `linting` tox env is deleted in 3c6b070, running `tox -e py3,linting` would just run unit tests twice.


nicoddemus:
Thanks!

PR #339, opened 2023-01-16
pylint and flake8 updates

marksmayo:
- cleaned up imports
 - added trailing commas (python 3.5+)
 - removed unused variables
 - fixed f-strings
 - black formatting


nicoddemus:
Thanks @marksmayo,

However I'm particularly not a big fan of pylint/flake8 and for that reason we do not have pre-commit rules set to ensure complience, so probably soon as the code evolves we will start violating flake8/pylint rules. 

Thanks again!

Issue #254, opened 2021-08-23
Patch methods should return AsyncMock if the target is an async method


GDYendell:
This would match the behaviour of the equivalent methods in [`mock`](https://github.com/python/cpython/blob/main/Lib/unittest/mock.py#L1687)

Currently, I can pass `new_callable=AsyncMock` to make it work, but then mypy type checking fails because the type hints state that the methods return `MagicMock`. The return type is a union in the [mock stubs](https://github.com/python/typeshed/blob/master/stubs/mock/mock.pyi#L170).

nicoddemus:
Hi @GDYendell, thanks for bringing this to attention.

nicoddemus:
Hi @GDYendell,

There's a PR for this in #364, but I wonder if you could post an example of a `mypy` failure? See my comments in https://github.com/pytest-dev/pytest-mock/pull/364#discussion_r1223710450.

GDYendell:
Hi @nicoddemus 

This was a while ago now so I doing some digging trying to remember. There is a chain of issues that leads to https://github.com/testing-cabal/mock/issues/500 where I included a mypy error.

The repo this was happening is https://github.com/dls-controls/odinprocservcontrol. I have created a `mypy` branch from an old commit to recreate the problem [here](https://github.com/dls-controls/odinprocservcontrol/actions/runs/5219900027/jobs/9422288095#step:6:60). It [works](https://github.com/dls-controls/odinprocservcontrol/actions/runs/5219807227/jobs/9422082840#step:6:87) on master with the [updated](https://github.com/dls-controls/odinprocservcontrol/commit/6056b44c96b0f946ebb07039e096782a888f8798) `types-mock`. If it is useful to recreate the problem locally, checkout the `mypy` branch and do a `pipenv install --dev --ignore-pipfile` and run `pipenv run mypy tests`.

I have briefly looked through #364, but don't immediately understand the detail of the change - is this useful information? Let me know if there is anything else I can do to help.

PR #364, opened 2023-06-07
Fix type annotation

egor43:
For issue #254 
unittest.patch return MagicMock or AsyncMock: https://github.com/python/cpython/blob/3.11/Lib/unittest/mock.py#L1467

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -112,7 +112,9 @@ class MockerFixture:
         else:
             raise ValueError("This mock object is not registered")
 
-    def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
+    def spy(
+        self, obj: object, name: str
+    ) -> Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]:
         """
         Create a spy of method. It will run method normally, but it is now
         possible to use `mock` call features with it, like call count.
@@ -244,7 +246,7 @@ class MockerFixture:
             autospec: Optional[object] = None,
             new_callable: object = None,
             **kwargs: Any
-        ) -> unittest.mock.MagicMock:
+        ) -> Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]:
             """API to mock.patch.object"""
             if new is self.DEFAULT:
                 new = self.mock_module.DEFAULT
@@ -273,7 +275,7 @@ class MockerFixture:
             autospec: Optional[builtins.object] = None,
             new_callable: builtins.object = None,
             **kwargs: Any
-        ) -> unittest.mock.MagicMock:
+        ) -> Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]:
             """This is equivalent to mock.patch.object except that the returned mock
             does not issue a warning when used as a context manager."""
             if new is self.DEFAULT:
@@ -301,7 +303,7 @@ class MockerFixture:
             autospec: Optional[builtins.object] = None,
             new_callable: Optional[builtins.object] = None,
             **kwargs: Any
-        ) -> Dict[str, unittest.mock.MagicMock]:
+        ) -> Dict[str, Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]]:
             """API to mock.patch.multiple"""
             return self._start_patch(
                 self.mock_module.patch.multiple,
@@ -343,7 +345,7 @@ class MockerFixture:
             autospec: Optional[builtins.object] = ...,
             new_callable: None = ...,
             **kwargs: Any
-        ) -> unittest.mock.MagicMock:
+        ) -> Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]:
             ...
 
         @overload
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -27,10 +27,16 @@ from ._util import parse_ini_boolean
 
 _T = TypeVar("_T")
 
-if sys.version_info[:2] > (3, 7):
+if sys.version_info >= (3, 8):
     AsyncMockType = unittest.mock.AsyncMock
+    MockType = Union[
+        unittest.mock.MagicMock,
+        unittest.mock.AsyncMock,
+        unittest.mock.NonCallableMagicMock,
+    ]
 else:
     AsyncMockType = Any
+    MockType = Union[unittest.mock.MagicMock, unittest.mock.NonCallableMagicMock]
 
 
 class PytestMockWarning(UserWarning):
@@ -112,9 +118,7 @@ class MockerFixture:
         else:
             raise ValueError("This mock object is not registered")
 
-    def spy(
-        self, obj: object, name: str
-    ) -> Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]:
+    def spy(self, obj: object, name: str) -> MockType:
         """
         Create a spy of method. It will run method normally, but it is now
         possible to use `mock` call features with it, like call count.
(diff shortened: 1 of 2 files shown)


egor43:
@nicoddemus Do you have time to take a look?

The-Compiler:
Looking at [what typeshed does](https://github.com/python/typeshed/blob/main/stdlib/unittest/mock.pyi#L275-L278), this seems to b a step in the right direction - though it breaks things on Python 3.7.

I suppose we could drop 3.7 support now, rather than introducing a conditional (given that [it's EOL](https://devguide.python.org/versions/) in 19 days)?

egor43:
I can do the same:
```python
if sys.version_info >= (3, 8):
    _Mock: TypeAlias = MagicMock | AsyncMock
else:
    _Mock: TypeAlias = MagicMock
```
Would such a solution work?

egor43:
Thank you @nicoddemus.
Sorry I didn't quite understand about using UNRELEASED for version. Could you tell me more?

nicoddemus:
The changes for each version starts with a header, for example `3.10.0 (2022-10-05)`, but the next version is not released yet, so just use `UNRELEASED`.

egor43:
I updated the changelog and made changes to the type annotation. Could you review the changes?

src/pytest_mock/plugin.py
```diff
@@ -112,7 +118,7 @@ def stop(self, mock: unittest.mock.MagicMock) -> None:
         else:
             raise ValueError("This mock object is not registered")
 
-    def spy(self, obj: object, name: str) -> unittest.mock.MagicMock:
+    def spy(self, obj: object, name: str) -> MockType:
```
nicoddemus, line 121: Thanks @egor43!

It seems to me though, by returning a Union, that will force users to check the return type before using it:

```python
foo_mock = mocker.spy(module, some_async_foo)
assert isinstance(foo_mock, mocker.AsyncMock)
# use foo_mock as mocker.AsyncMock.
```

If the above is right, I don't think we need this PR, because that should work regardless -- however I think it will make mypy consider the code below it unreachable...

Possibly a better way to type annotate this would be to use `@overload`s, with different return types based on the input object, something like this:

https://github.com/python/typeshed/blob/852ac3efb3da704267082cc611557214f2f87aac/stubs/mock/mock/mock.pyi#L244-L270C17

As before thanks for the PR, but I wanted to discuss this to ensure we are getting the correct fix.

nicoddemus: Btw using `@overloads` here seems very complicated, given the really dynamic nature of `mock` -- I was not asking you to do it,  I was just commenting on a possibility, and see if I might be missing something.

nicoddemus: Did some testing:

```python
import unittest.mock
from pytest_mock import MockerFixture

class Foo:
    def normal_func(self) -> None:
        ...

    async def async_func(self) -> None:
        ...

async def test_x(mocker: MockerFixture) -> None:
    m = mocker.patch.object(Foo, "normal_func")
    reveal_type(m)
    await m()
    m.assert_called()
    assert m.await_count == 1

async def test_y() -> None:
    with unittest.mock.patch.object(Foo, "normal_func") as m:
        reveal_type(m)
        await m()
        m.assert_called()
        assert m.await_count == 1
```

Running `mypy` in `main`:

```
λ mypy .tmp\test_foo.py
.tmp\test_foo.py:13: note: Revealed type is "unittest.mock.MagicMock"
.tmp\test_foo.py:20: note: Revealed type is "Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]"
Success: no issues found in 1 source file
```

So right of the gate I'm surprised here: I expected `mypy` to complain in `test_x` about being awaited and `m.await_count`, given those attribute are from `AsyncMock` and do not exist in `MagicMock`. However of course this is the nature of `mock`, it will automatically create any attribute when accessed.

Running this on your branch:

```
λ mypy .tmp\test_foo.py
.tmp\test_foo.py:13: note: Revealed type is "Union[unittest.mock.MagicMock, unittest.mock.AsyncMock, unittest.mock.NonCallableMagicMock]"
.tmp\test_foo.py:20: note: Revealed type is "Union[unittest.mock.MagicMock, unittest.mock.AsyncMock]"
Success: no issues found in 1 source file
```

So the type annotation in `test_x` is now "correct" -- I use "correct" in quotes because as far as I understand, `mypy` cannot really help here, because it does not matter if `mocker.patch` returns `MagicMock` or `MagicMock | AsyncMock`, because it does not matter what attribute is accessed, it must always succeed.

I wonder what was failing in #254.

-----

So if my understand is correct, unfortunately the new type annotations does not seem to be helpful at all in catching errors.

Having said all that, I think it is harmless to merge the PR, at least we will be better align with how `mock` is typed.

nicoddemus: I've asked @GDYendell for clarification in #254.

egor43: Thanks for the in-depth analysis of the issue and for the flaw found.
Yes, I was thinking about using ```@overloads``` . But it really seemed to me a very difficult decision. In addition, ```unittest.mock._patch``` does a lot of runtime checks to infer the final type.

egor43: You're right. mypy does not react in any way to an attempt to access non-existent attributes.
But some IDEs like VSCode highlight the attribute as unknown. Because of this, you cannot jump to the implementation of the method.
My current pull request returns a more truthy type that both mypy and the IDE can work with.

nicoddemus: Thanks @GDYendell, as per information you provided in #254, I managed to reproduce the issue you had using this code:

```python
from unittest.mock import AsyncMock

from pytest_mock import MockerFixture


class Control:

    async def _start_processes(self) -> None:
        ...

    async def start_processes(self, x: int) -> None:
        ...


async def test_start_processes_press(
        control: Control, mocker: MockerFixture
) -> None:
    mocker.patch.object(control, "_start_processes", new_callable=AsyncMock)
    await control.start_processes(1)
    control._start_processes.assert_awaited_once_with()
```

This gives the following errors, both in this PR and `main`:

```
 λ mypy .tmp\test_foo2.py
.tmp\test_foo2.py:20: error: "Callable[[], Coroutine[Any, Any, None]]" has no attribute "assert_awaited_once_with"
[attr-defined]
        control._start_processes.assert_awaited_once_with()
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 1 source file)
```

Which is the same error you had when you reported #254.

But the problem here is how it is written: `mypy` can't know that `control._start_processes` has been mocked and is now a `MagicMock`, so it correctly points the problem that a callable does not have the `assert_awaited_once_with` method.

The fix is to simply use the `MagicMock` object returned by `mocker.patch.object`:

```python
async def test_start_processes_press(
        control: Control, mocker: MockerFixture
) -> None:
    m = mocker.patch.object(control, "_start_processes", new_callable=AsyncMock)
    await control.start_processes(1)
    m.assert_awaited_once_with()
```

So this passes, again both in this PR and in `main`.

It was fixed in that repository in https://github.com/dls-controls/odinprocservcontrol/commit/6056b44c96b0f946ebb07039e096782a888f8798 by doing essentially the same thing, but using `unittest.mock` directly:

```python
async def test_start_processes_press(
        control: Control
) -> None:
    with unittest.mock.patch.object(control, "_start_processes", new_callable=AsyncMock) as m:
        await control.start_processes(1)
        m.assert_awaited_once_with()
```

In conclusion, the problem was never the type annotation. :sweat_smile:

I think however given that this PR does improve IDE auto-completion, I see no harm merging it.

Thanks everyone!

nicoddemus (approved):

Issue #386, opened 2023-10-06
Not sure why this is not working


jiz17043:
![image](https://github.com/pytest-dev/pytest-mock/assets/44070470/0f514a48-fdbb-462a-9bbf-c61747a1b46b)
![image](https://github.com/pytest-dev/pytest-mock/assets/44070470/ad059f8e-5a0c-4d10-80db-0c23bd212452)
Try to mock the response of the function in the httprequest_gecko module but it seems not working. Please let me know if you have any suggestions. Thank you!
I am using the latest versions of pytest (7.1.2) and pytest-mock (3.11.1) on Python 3.9.2.

The-Compiler:
Not sure off hand what the issue is, but looks like you're not even using pytest-mock in that screenshot - so this isn't an issue with the plugin.

jiz17043:
I would like to ask for any suggestions if I want to mock a function.

nicoddemus:
@jiz17043 see this section in the docs, as it is likely relevant:

https://docs.python.org/3/library/unittest.mock.html#where-to-patch

PR #387, opened 2023-10-06
Add "where to patch" link to the docs

nicoddemus:
As #386 shows, this is a common problem, so add a link to the docs to help users.


The-Compiler:
Sounds good! Maybe also link this somewhere? https://nedbatchelder.com/blog/201908/why_your_mock_doesnt_work.html

nicoddemus:
> Sounds good! Maybe also link this somewhere?

Good idea, done!

PR #390, opened 2023-10-15
Add support for `resetall` to `create_autospec`

inikolaev:
This PR adds support for `resetall` to `create_autospec`.

1. I have added a new `create_autospec` function to the `MockerFixture`, that wraps original `create_autospec` and adds a reference to the `_patches_and_mocks` list, so it could be reset by `resetall`.

2. `stopall` had to be modified in order to ignore `None` patcher, maybe there's a better way to do this instead.

3. Added new test cases to `test_mocker_resetall` test
4. `test_mocker_aliases` is failing - maybe we don't need to test it for `create_autospec` anymore, since we have it tested in `test_mocker_resetall`?

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -66,11 +66,19 @@ class MockerFixture:
         self.call = mock_module.call
         self.ANY = mock_module.ANY
         self.DEFAULT = mock_module.DEFAULT
-        self.create_autospec = mock_module.create_autospec
         self.sentinel = mock_module.sentinel
         self.mock_open = mock_module.mock_open
         if hasattr(mock_module, "seal"):
             self.seal = mock_module.seal
+
+    def create_autospec(
+        self, spec: Any, spec_set: bool = False, instance: bool = False, **kwargs: Any
+    ) -> MockType:
+        m: MockType = self.mock_module.create_autospec(
+            spec, spec_set, instance, **kwargs
+        )
+        self._patches_and_mocks.append((None, m))
+        return m
 
     def resetall(
         self, *, return_value: bool = False, side_effect: bool = False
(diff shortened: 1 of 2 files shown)


tests/test_pytest_mock.py
```diff
@@ -185,23 +194,35 @@ def test_mocker_resetall(mocker: MockerFixture) -> None:
...
 
     mocker.resetall(return_value=True, side_effect=True)
 
     assert isinstance(listdir.return_value, mocker.Mock)
     assert open.side_effect is None
 
+    if sys.version_info >= (3, 9):
+        # The reset on child mocks have been implemented in 3.9
+        # https://bugs.python.org/issue38932
+        assert mocked_object.run.return_value != "mocked"
```
inikolaev, line 224: Don't really like this check here, maybe test needs to be split, and then it's easier to use `pytest.mark.skipif` to skip for 3.8 and below.

inikolaev: Or maybe have a completely separate test for resetting child mocks

nicoddemus:
Thanks @inikolaev for the patch, appreciate it!

> stopall had to be modified in order to ignore None patcher, maybe there's a better way to do this instead.

Seems good enough. 👍  

> test_mocker_aliases is failing - maybe we don't need to test it for create_autospec anymore, since we have it tested in test_mocker_resetall?

Yes. 👍

src/pytest_mock/plugin.py
```diff
@@ -102,7 +110,8 @@ def stopall(self) -> None:
         times.
         """
         for p, m in reversed(self._patches_and_mocks):
-            p.stop()
+            if p:
```
nicoddemus, line 113: Let's be more explicit here:

```suggestion
            if p is not None:
```

Given `p` is Any, is theoretically possible for it to implement `__bool__` and return `False` for some reason.

tests/test_pytest_mock.py
nicoddemus: The check is fine, when we drop support for Python versions it is common to search for `sys.version_info` checks like these. 👍

nicoddemus (changes requested):
Thanks @inikolaev, looks great so far!

Please take a look at my comments.

PR #395, opened 2023-10-31
Ship py.typed with the source distribution

adamtheturtle:
Without this, I was hitting mypy errors that there were no stub found for pytest-mock, as I had not installed the wheel.

See
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages for mypy documentation which describes this.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+global-include *.typed


nicoddemus:
Hi @adamtheturtle, 

Thanks, but the `py.typed` file is already included via:

https://github.com/pytest-dev/pytest-mock/blob/65b40f421db7ff67270123168de3fde5f84b65f8/setup.py#L10-L12

It is also verifiable that the `py.typed` file is included in the distribution, here is the latest source dist downloaded from PyPI:

image

So if you are having trouble the cause must be something else.

adamtheturtle:
Thank you @nicoddemus . I'll create an issue with clear reproduction steps.

adamtheturtle:
The error I saw was because of an issue from me. I'm still confused about the discrepancy between https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages and what we're seeing here.

That explains:

```
The instructions above are enough to ensure that the built wheels contain the appropriate files. However, to ensure inclusion inside the sdist (.tar.gz archive), you may also need to modify the inclusion rules in your MANIFEST.in:
```

nicoddemus:
Not sure what you mean exactly, can you clarify? I checked and `py.typed` is present in both the sdist and wheel files.

adamtheturtle:
@nicoddemus I believe you have proved that everything is working fine with just the `package_data` setting.

The mypy documentation describes the `package_data` option, but it seems to explain that this "may" not be "enough" for `sdist`s, and that `MANIFEST.in` changes may also be needed. I'm not sure the circumstances of "may".

nicoddemus:
Got it, perhaps if it is built by an old `setuptools` or something like that.

PR #399, opened 2023-12-06
autospec for spy: remove if else which was in place for python < 3.7

yesthesoup:
I made this PR in the past #224 which worked on an `if/else` for the `spy` feature to get around a bug in python that was fixed in 3.7. Now that this library supports only >= 3.8 this check is no longer needed.

I suppose this would be "breaking changes" for anyone who may be incorrectly calling methods on class/static method spies?

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -137,14 +137,6 @@ class MockerFixture:
         :return: Spy object.
         """
         method = getattr(obj, name)
-        if inspect.isclass(obj) and isinstance(
-            inspect.getattr_static(obj, name), (classmethod, staticmethod)
-        ):
-            # Can't use autospec classmethod or staticmethod objects before 3.7
-            # see: https://bugs.python.org/issue23078
-            autospec = False
-        else:
-            autospec = inspect.ismethod(method) or inspect.isfunction(method)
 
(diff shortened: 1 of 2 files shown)


yesthesoup:
> Hey @yesthesoup, thanks for the follow up!
> 
> > I suppose this would be "breaking changes" for anyone who may be incorrectly calling methods on class/static method spies?
> 
> To confirm, would any Python 3.8+ user be affected by this change in any way?

I believe they would only be affected if their tests were silently passing when they shouldn't be like in the examples in the Python docs:

https://docs.python.org/3/library/unittest.mock.html#auto-speccing

```python
mock = Mock(name='Thing', return_value=None)
mock(1, 2, 3)
mock.assret_called_once_with(4, 5, 6)  # would not throw error in their current tests - silently pass
```

e.g. if they had some class method / static method spies that pytest-mock was previously creating with `autospec=false` due to the older versions' bug *and* those spies had incorrectly typed methods called on them

Should be a very small percentage of users, maybe none, and if this is merged, it would probably be good for those affected to update and have those silent failures fixed.

nicoddemus:
I see, thanks! I guess we might leave out this from the CHANGELOG then for now, and if somebody is affected by this, we can add a note accordingly.

Thanks again!

PR #410, opened 2024-02-12
pyproject.toml: Move metadata here from setup.cfg

mkoeppe:
Prepared with the help of `pdm init`


nicoddemus:
Thanks @mkoeppe!

Went ahead and moved everything off `setup.py`. Thanks for the nudge.

PR #466, opened 2024-10-13
Adds support to python 3.13

jairhenrique:


.github/workflows/test.yml
```diff
@@ -63,6 +63,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python }}
+        allow-prereleases: true
```
The-Compiler, line 66: Is this needed? Python 3.13 should not be a prerelease anymore?

jairhenrique: @The-Compiler to Python 3.13 no, Github already release the final version, but it could be useful if we want to prepare the ci/code to next version!

In the manifest 3.14 is now available https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json.

Do you think it's better to take it off?

nicoddemus: I think it is published now, so we can take it off? I'm doing it to see.

.github/workflows/test.yml
```diff
@@ -25,7 +25,8 @@ jobs:
       - name: Build and Check Package
         uses: hynek/build-and-inspect-python-package@v2.9.0
         with:
-          attest-build-provenance-github: 'true'
+          # Disabling because this is failing currently, see #466.
```
webknjaz, line 28: @nicoddemus https://github.com/hynek/build-and-inspect-python-package/issues/105#issuecomment-2136262019

nicoddemus: To summarize, it is better then to use `with attestations` in the publish-pypi action instead of in `hynek/build-and-inspect-python-package`?

webknjaz: Yeah, that's my stance. Though, not everybody feels that way.

FWIW, GitHub's own attestations are embedded into the UI. The PyPI ones are upload provenance, not necessarily build provenance, technically.

Issue #468, opened 2024-10-24
DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated


hroncok:
Hello,

when building dependent packages in Fedora with Python 3.14.0a1, I got this when building meson-python:

```
____________________________ test_invalid_build_dir ____________________________

package_pure = PosixPath('/builddir/build/BUILD/python-meson-python-0.17.0-build/meson_python-0.17.0/tests/packages/pure')
tmp_path = PosixPath('/tmp/pytest-of-mockbuild/pytest-0/test_invalid_build_dir0')
mocker = <pytest_mock.plugin.MockerFixture object at 0x7f1b7d91b750>

    @pytest.mark.skipif(
        sys.version_info < (3, 8),
        reason="unittest.mock doesn't support the required APIs for this test",
    )
    def test_invalid_build_dir(package_pure, tmp_path, mocker):
>       meson = mocker.spy(mesonpy.Project, '_run')

/builddir/build/BUILD/python-meson-python-0.17.0-build/meson_python-0.17.0/tests/test_project.py:168: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.14/site-packages/pytest_mock/plugin.py:199: in spy
    if asyncio.iscoroutinefunction(method):
/usr/lib64/python3.14/asyncio/coroutines.py:23: in iscoroutinefunction
    warnings._deprecated("asyncio.iscoroutinefunction",
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'asyncio.iscoroutinefunction'
message = '{name!r} is deprecated and slated for removal in Python {remove}; use inspect.iscoroutinefunction() instead'

    def _deprecated(name, message=_DEPRECATED_MSG, *, remove, _version=sys.version_info):
        """Warn that *name* is deprecated or should be removed.
    
        RuntimeError is raised if *remove* specifies a major/minor tuple older than
        the current Python version or the same version but past the alpha.
    
        The *message* argument is formatted with *name* and *remove* as a Python
        version tuple (e.g. (3, 11)).
    
        """
        remove_formatted = f"{remove[0]}.{remove[1]}"
[...]

PR #469, opened 2024-10-24
Replace asyncio.iscoroutinefunction with inspect.iscoroutinefunction

hroncok:
DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead

Fixes https://github.com/pytest-dev/pytest-mock/issues/468

### DIFF
--- a/src/pytest_mock/plugin.py
+++ b/src/pytest_mock/plugin.py
@@ -1,4 +1,3 @@
-import asyncio
 import builtins
 import functools
 import inspect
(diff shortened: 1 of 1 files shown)


hroncok:
I still see the warning here when I run `tox -e py314` because of pytest-asyncio. It is fixed there in https://github.com/pytest-dev/pytest-asyncio/commit/d0436985afa086a01544545ff6c38bd7383e25a9 but not yet released.

---

Also, about the CI failure... no idea what it means.

webknjaz:
@seifertm could you make a release if `pytest-asyncio`? Other plugins like `pytest-mock` start testing against Python 3.14 and hit that warning fixed in the main branch.

nicoddemus:
The failure is something related to `hynek/build-and-inspect-python-package`, because we are having the same error in https://github.com/pytest-dev/pytest-mock/pull/466, which ran fine two weeks ago.

nicoddemus:
@hroncok rebased on `main`, which fixed the build-and-check-package problem.

PR #471, opened 2024-11-01
Add `AsyncMock` to list of available modules from `unittest.mock` in documentation

mishaschwartz:
When using `unittest.mock` as the `mock_module`, the `AsyncMock` class has been made accessible directly from the `mocker` fixture since v3.2.0:

https://github.com/pytest-dev/pytest-mock/blob/9b6e10645f6f15b6972391fc4295faa1f658c1c3/src/pytest_mock/plugin.py#L50-L51

I noticed that `AsyncMock` is not included in the list in the documentation so this PR adds it in.


nicoddemus (approved):
Thanks!

PR #503, opened 2025-05-12
Add support to Python 3.14

edgarrmondragon:


tox.ini
```diff
@@ -1,13 +1,14 @@
...
-envlist = py{38,39,310,311,312,313}, norewrite, pytest6
+envlist = py{38,39,310,311,312,313,314}, norewrite, pytest6
 
 [testenv]
 deps =
     coverage
     mock
     pytest-asyncio
     pytest6: pytest==6.2.5
+pip_pre = true
```
edgarrmondragon, line 11: https://github.com/pytest-dev/pytest-asyncio/issues/1025#issuecomment-2866587131

nicoddemus: I think we can wait a bit until this is no longer necessary?

nicoddemus (commented):
Thanks!

tox.ini
edgarrmondragon: Yeah, totally 👍

seifertm: Support for the development versions of Python 3.14 is now available in pytest-asyncio v1.0.0 :)

nicoddemus, line 11:
```suggestion
```

nicoddemus: Thanks @seifertm for the ping.

PR #524, opened 2025-09-01
Add `spy_return_iter` attribute to `spy`

vokracko:
Recently I've run into situation where I wanted to `spy` on a method returning `Iterator`.
That wasn't possible because the iterator is already consumed by the time you try to assert it.

In this PR I add a new attribute, `spy_return_iter` storing a duplicate of the returned iterator (leveraging `itertools.tee`).

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -81,6 +81,7 @@ are available (like ``assert_called_once_with`` or ``call_count`` in the example
 In addition, spy objects contain two extra attributes:
 
 * ``spy_return``: contains the last returned value of the spied function.
+* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator.
 * ``spy_return_list``: contains a list of all returned values of the spied function (new in ``3.13``).
 * ``spy_exception``: contain the last exception value raised by the spied function/method when
   it was last called, or ``None`` if no exception was raised.
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -81,7 +81,7 @@ are available (like ``assert_called_once_with`` or ``call_count`` in the example
 In addition, spy objects contain two extra attributes:
 
 * ``spy_return``: contains the last returned value of the spied function.
-* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator.
+* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator. Uses `tee <https://docs.python.org/3/library/itertools.html#itertools.tee>`__) to duplicate the iterator.
 * ``spy_return_list``: contains a list of all returned values of the spied function (new in ``3.13``).
 * ``spy_exception``: contain the last exception value raised by the spied function/method when
   it was last called, or ``None`` if no exception was raised.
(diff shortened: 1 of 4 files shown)


tests/test_pytest_mock.py
```diff
@@ -521,9 +541,67 @@ def __call__(self, x):
...
+
+    foo = Foo()
+    spy = mocker.spy(foo, "bar")
+    result = list(foo.bar())
+
+    assert result == [0, 1, 2]
+    assert spy.spy_return is not None
+    assert spy.spy_return_iter is not None
+    assert list(spy.spy_return_iter) == result
+    assert spy.spy_return_list == [ANY]
```
nicoddemus, line 562: Does this work?

```suggestion
    assert isinstance(spy.spy_return_list[0], Iterator)
```

tests/test_pytest_mock.py
```diff
@@ -521,9 +541,67 @@ def __call__(self, x):
...
+    foo = Foo()
+    spy = mocker.spy(foo, "bar")
+    result_iterator = list(foo.bar())
+
+    assert result_iterator == [0, 1, 2]
+    assert list(spy.spy_return_iter) == result_iterator
+
+    result_iterable = foo.bar()
+    assert result_iterable == [3, 4, 5]
+    assert spy.spy_return_iter is None
```
nicoddemus, line 602: Minor suggestion to make this clearer:


```suggestion
def test_spy_return_iter_resets(mocker: MockerFixture) -> None:
    class Foo:
        iterables = [
            (i for i in range(3)),
            99,
        ]

        def bar(self) -> Iterable[int]:
            return self.iterables.pop(0)

    foo = Foo()
    spy = mocker.spy(foo, "bar")
    result_iterator = list(foo.bar())

    assert result_iterator == [0, 1, 2]
    assert list(spy.spy_return_iter) == result_iterator

    assert foo.bar() == 99
    assert spy.spy_return_iter is None
```

Mixing iterator and iterables on this test is not necessary for the logic and adds a small bit of confusion.

docs/usage.rst
```diff
@@ -81,6 +81,7 @@ are available (like ``assert_called_once_with`` or ``call_count`` in the example
 In addition, spy objects contain two extra attributes:
 
 * ``spy_return``: contains the last returned value of the spied function.
+* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator.
```
nicoddemus, line 84:
```suggestion
* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator. Uses `tee <https://docs.python.org/3/library/itertools.html#itertools.tee>`__) to duplicate the iterator.
```

nicoddemus (changes requested):
Thanks for the contribution @vokracko!

Left some minor comments, please take a look.

nicoddemus:
I will merge it shortly, nevermind the error.

nicoddemus:
Thanks @vokracko!

tests/test_pytest_mock.py
vokracko: I had a fix ready for this but you were faster

Issue #529, opened 2025-09-05
itertools._tee object has no attribute error with 3.15.0


kasium:
With the newest version (3.15.0) the below code fails with `AttributeError: 'itertools._tee' object has no attribute 'start'`

foo.py
```
class Client:
    def start(self):
        pass

class Lock:
    def create_client(self):
        return Client()
    def start(self):
        self._client = self.create_client()
        self._client.start()

```

test_foo.py
```py
import foo

def test(mocker):
    mocker.patch("foo.Client")
    lock = foo.Lock()
    mocker.spy(lock, "create_client")
    lock.start()
```

The problem ist, that spy checks if the returned instance is an iterator, but each `MagicMock` is an iterator which leads to issues.
The logic should exclude any mock object IMO which should solve the issue

nicoddemus:
Thanks @kasium for the report!

@vokracko could you take a look at this?

PR #537, opened 2025-09-16
Set `spy_return_iter` only when explicitly requested

vokracko:
Fixes https://github.com/pytest-dev/pytest-mock/issues/529

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -78,10 +78,10 @@ also tracks function/method calls, return values and exceptions raised.
 The object returned by ``mocker.spy`` is a ``MagicMock`` object, so all standard checking functions
 are available (like ``assert_called_once_with`` or ``call_count`` in the examples above).
 
-In addition, spy objects contain two extra attributes:
+In addition, spy objects contain four extra attributes:
 
 * ``spy_return``: contains the last returned value of the spied function.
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -78,10 +78,10 @@ also tracks function/method calls, return values and exceptions raised.
 The object returned by ``mocker.spy`` is a ``MagicMock`` object, so all standard checking functions
 are available (like ``assert_called_once_with`` or ``call_count`` in the examples above).
 
-In addition, spy objects contain two extra attributes:
+In addition, spy objects contain four extra attributes:
 
 * ``spy_return``: contains the last returned value of the spied function.
-* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator. Uses `tee <https://docs.python.org/3/library/itertools.html#itertools.tee>`__) to duplicate the iterator.
+* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator and spy was created using ``.spy(..., duplicate_iterators)``. Uses `tee <https://docs.python.org/3/library/itertools.html#itertools.tee>`__) to duplicate the iterator.
```
nicoddemus, line 84:
```suggestion
* ``spy_return_iter``: contains a duplicate of the last returned value of the spied function if the value was an iterator and spy was created using ``.spy(..., duplicate_iterators=True)``. Uses `tee <https://docs.python.org/3/library/itertools.html#itertools.tee>`__) to duplicate the iterator.
```

nicoddemus (approved):

kasium:
Thanks a lot!

PR #593, opened 2026-07-16
Allow mocker.spy on frozen dataclass attributes

### LINKED ISSUE #280: spy on frozen dataclass
Spy does not work on frozen dataclasses, it would be nice if it did (for instance, by using `object.__setattr__(obj, name, spy)`

```python
from dataclasses import dataclass
from typing import Callable


def noop(*args, **kwargs):
    pass


@dataclass(frozen=True)
class MyFrozenClass:
    something: Callable = noop


def test_spy_frozen_dataclass(mocker: MockerFixture):
    o = MyFrozenClass()
    spy = mocker.spy(o, 'something')

```

pctablet505:
Closes #280

`mocker.spy` installs its wrapper through `mock.patch.object`, which sets
the attribute with plain `setattr`/`delattr`. Frozen dataclasses override
both to always raise `FrozenInstanceError`, so spying on one of their
fields never worked, as reported in the issue.

This temporarily relaxes `__setattr__`/`__delattr__` on the frozen
dataclass's class (via `object.__setattr__`/`__delattr__`, the same
workaround mentioned in the issue) only while the spy is being installed
and later while it is stopped/restored, so the instance is frozen again
as soon as the mocker is done with it. Non-frozen objects go through the
exact same code path as before.

Added a test covering spying on a frozen dataclass attribute, calling
through the spy, and confirming the instance is still frozen after
`stopall()`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Releases
 Unreleased
 ----------
 
+* `#280 <https://github.com/pytest-dev/pytest-mock/issues/280>`_: ``mocker.spy`` now works on attributes of frozen dataclass instances.
 * `#547 <https://github.com/pytest-dev/pytest-mock/issues/547>`_: Added ``SpyType`` for annotating ``mocker.spy`` results.
 * Dropped support for EOL Python 3.9.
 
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks, but due to the complexity and the fact that `mocker.patch.object` does not support it, pytest-mock won't support it either.

PR #601, opened 2026-08-19
Support spying on properties

### LINKED ISSUE #35: Spying on properties
I've tried to use `spy` on object properties, but it fails with `AttributeError: can't set attribute` in `unittest/mock.py`.

I've started looking into adding support for (through `PropertyMock`), but failed to make it work without accessing/calling the property method during setup of the spy - it would be better to only call it when it's accessed from the test.

I would appreciate any feedback / help in this regard.

ChrisJr404:
This teaches `mocker.spy` to work on properties, which currently fails with `AttributeError: can't set attribute` because spy reads the value with `getattr` and then tries to patch the attribute on the instance.

The fix detects when the named attribute is a `property` and spies on its getter instead. Reads still return the real, computed value, and the returned spy records each access along with `spy_return`, `spy_return_list` and `spy_exception`, just like spying on a method. The original setter and deleter are left untouched. Because a property lives on the class, the spy applies to every instance until it is undone, which I noted in the docs.

Closes #35. There's a changelog entry and tests covering instance and by-class spying, exceptions, the setter staying intact, and the property being restored after `stopall`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Releases
 Unreleased
 ----------
 
+* `#35 <https://github.com/pytest-dev/pytest-mock/issues/35>`_: ``mocker.spy`` now works on properties, spying on the getter while still returning the real value.
 * `#547 <https://github.com/pytest-dev/pytest-mock/issues/547>`_: Added ``SpyType`` for annotating ``mocker.spy`` results.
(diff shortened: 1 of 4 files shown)


nicoddemus:
Given this is not directly supported by unittest.mock, I would not like to introduce this into pytest-mock itself. Thanks though!

PR #602, opened 2026-08-21
Add support for Python 3.15

kytta:
Python 3.15.0 rc1 is out, which means that there will be no more ABI changes, and the API is pretty much stable. It is now safe to test against and to specify support for it :)

This PR adds Python 3.15 support to pytest-mock by enabling testing on it and adding a PyPI classifier.


nicoddemus:
Thanks!

PR #607, opened 2026-09-07
fix(plugin): ensure mocker.stub is not detected as a coroutine function

### LINKED ISSUE #375: mocker.stub() is a coroutine function
```
def test_icf_sync(mocker):
    func = mocker.stub('func')
    assert not inspect.iscoroutinefunction(func)


def test_icf_async(mocker):
    func = mocker.async_stub('func')
    assert inspect.iscoroutinefunction(func)
```

Expected: `mocker.stub()` returns something that doesn't resemble a coroutine.

Observed: `mocker.stub()` is detected as a coroutine.

CyberneticX-Tech:
Resolves #375

Because `mocker.stub()` constructs a `MagicMock(spec=lambda *args, **kwargs: None)`, `isinstance(stub, types.FunctionType)` evaluates to `True`. When `inspect.iscoroutinefunction(stub)` is called, it inspects `stub.__code__.co_flags & CO_COROUTINE`. Accessing `stub.__code__` creates a child `MagicMock`, whose bitwise `&` operation also returns a `MagicMock`, which evaluates to truthy, incorrectly marking the synchronous stub as a coroutine function.

Assigning `fn.__code__` to `stub.__code__` ensures `stub` retains the real code flags of a standard synchronous function while preserving all other `MagicMock` behaviors.


nicoddemus:
Not sure we want to extend stub to be more "clever" than a plain `MagicMock`, but thanks anyway!

PR #609, opened 2026-09-08
Preserve assertion failures for keyword call expectations

oyeong011:
Calling `stub.assert_has_calls(calls=[expected])` after an unexpected call raises `IndexError` instead of `AssertionError`. The upstream `unittest.mock` signature accepts the `calls` keyword. Read expected calls from that keyword when no positional argument was supplied.

The regression fails with IndexError before the change and passes afterward. The full suite passes: 92 tests on macOS/Python 3.12.13 with pytest 9.1.1. Ruff check/format and diff checks pass. Validation used installed wheel metadata for pytest assertion rewriting while PYTHONPATH pointed to the source checkout; no tests were excluded. Other platform/version combinations were not run.

AI assistance: OpenAI Codex assisted with implementation, tests, and this description. The reported checks were executed locally.


src/pytest_mock/plugin.py
nicoddemus, line 526: Can we instead just declare the actual signature?

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls

PR #612, opened 2026-09-20
docs: import os in the opening usage example

agammann:
The opening usage example calls `mocker.patch.object(os, "listdir", autospec=True)` without importing `os`, so copying it into a test module raises `NameError`.

Add the import to make the example runnable as shown.

Validation: extracted the original and corrected snippets into temporary test modules and ran them with pytest and this checkout's `mocker` fixture. The original fails with `NameError: name 'os' is not defined`; the corrected example passes. The strict Sphinx HTML build also passes.


nicoddemus:
Thanks

